PR #84, opened 2020-02-19
Cell wrapping

notsidney:
Wrap text in long text and rich text cells


notsidney:
Also: removed the popup to edit rich text, long text, and JSON since there’s a bug with the HTML editor: you can’t add/edit links in the popup view. Now when double-clicking a cell, it will open the side drawer and scroll to the correct field

PR #93, opened 2020-03-05
Temporarily disable side drawer due to performance issues on main table

notsidney:


harinij:
Cannot be merged right now as longtext and richtext will not work. @notsidney is looking into alternative options @shamsmosowi

PR #183, opened 2020-09-03
store sidedrawer docRef in the url

AntlerEngineering:


notsidney (approved):
✅ Should work as expected

PR #220, opened 2020-09-29
New Table - Copy from existing schema

matthewtsmith:
This PR contains two new features:

### Support creating a new table by copying from an existing table schema

When creating a table we often want to copy a schema from an already existing table rather than reconfiguring the columns. To support this, I added the ability in the `TableSettingsDialog` to specify another schema to use when initializing the new Firetable schema.

Screen Shot 2020-09-29 at 8 44 27 AM
Screen Shot 2020-09-29 at 8 44 32 AM

### Customize the data path prefix

We have an existing firestore database that we are connecting firetable to. We do not want to litter the root level of our firestore db with a bunch of the firetable collections. To avoid this, I added a new environment variable `REACT_APP_DATA_PATH_PREFIX` that allows you to specify a path to store the data in. In our case, we have it set to

REACT_APP_DATA_PATH_PREFIX = _FIRETABLE_/data

By setting this, the collection containing the data will be in `_FIRETABLE/data/{collection_name}`


shamsmosowi:
hi @matthewtsmith thanks again for your contributions, just to clarify you're aware that putting your data in subcollections in firestore loses the basic sorting and filtering functionality that is provided by default, and you'll have to create an index of each column.

matthewtsmith:
@shamsmosowi I'm not aware of that limitation and in testing this change I did not find that to be the case. No indexes were required in our firestore after moving our data into a subcollection. I also could not find any documentation to support your statement. Do you have more information you can send me regarding this?

Regardless, this is an environment flag that is not enabled by default so I think it's safe to include as a user would have to actively seek it out to turn it on. With that said, the primary purpose of this PR is to enable schema copying so if you are not comfortable with the subcollection support I can pull that out of this PR. Let me know your thoughts.

shamsmosowi:
I think i might have confused it with the collection group limitation🤔
I will test it out and merge it this week
Thanks!

shamsmosowi:
hi @matthewtsmith ,
We've reimplemented the copying table schema feature in #271 , feel free to open a separate PR for just the custom data path prefix feature, would be happy to merge

PR #259, opened 2020-11-04
update helper text in create table form

zdebra:
Hello, I am setting up Firetable for the first time and I was cracking my head around why my tables are not visible in the side menu. I think that the helper text is misleading as I was passing user email instead of role to this field. 

I propose to change `users` to `roles` which is more appropriate.


notsidney:
Hi @zdebra, thanks for your contribution! This is definitely one of the trickier parts of setup, so if there is any way we can further improve our documentation with roles and security rules, please let us know.

notsidney (approved):

PR #345, opened 2021-03-11
Feature/cloud run

bofeiw:


ft_build/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export async function validateSparks(
+  sparks: string,
+  user: admin.auth.UserRecord
+) {
+  let parsedSparks;
+
+  // try to parse sparks
+  try {
+    parsedSparks = eval(sparks);
```
shamsmosowi, line 70: try not to use eval in this cloud run instance, as it can provide surface for an injection attacks

shamsmosowi (approved):

PR #418, opened 2021-04-21
add force refresh derivatives/sparks to table header

scotty595:
Add a force refresh button to update the _ft_forcedUpdateAt time for all documents in a collection resulting in the sparks (and derivatives when combined with my previous push) refreshing for all documents.

Force refresh button will be hidden if the user is not admin or the table contains no sparks or derivatives

Example:
**Step 1: Adding a new derivative:**
https://www.loom.com/share/8f5289495a074217a9d384bceff569e1

**Step 2: Force Refresh:**
https://www.loom.com/share/e1f831bdb3ba43dbb5fc398e2a1e3336

### DIFF
--- a/www/src/components/Table/TableHeader/ForceRefresh.tsx
+++ b/www/src/components/Table/TableHeader/ForceRefresh.tsx
@@ -1,0 +1,97 @@
+import React, { useState } from "react";
+
+import TableHeaderButton from "./TableHeaderButton";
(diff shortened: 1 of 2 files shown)


shamsmosowi:
Thanks Scotty! , 
Further development for controlling which derivatives and sparks will be refreshed/trigger would be ideal, This might be possible to implement by creating a callable version of the current FT functions. Let me know if you want give it a go, otherwise I might implement it myself

PR #500, opened 2021-08-27
allow user to clear column config selection in create table

qnkhuat:
When creating a table user can choose to copy column config. 
The previous version doesn't allow users to clear this selection once they select. 
This PR adds a clear button to fix that

Screen Shot 2021-08-27 at 15 07 05

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,12 @@
 </h3>
 <br/>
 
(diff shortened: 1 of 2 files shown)


shamsmosowi:
thanks @qnkhuat !

Issue #583, opened 2021-11-16
Import CSV Issues


lafronzt:
**Describe the bug**
When Importing a CSV, I would have expected it to generate a new row for each item. However, it keeps replacing the same row with the latest data rather than creating a new row.

**To Reproduce**
1. Create New table
2. Add the Columns you would like
3. Import CSV
4. See how there is only one row.

**Expected behavior**
I expected a new row for each item in the CSV

**Desktop (please complete the following information):**
 - OS: MacOS X
 - Browser: Chrome 
 - Version: Version 95.0.4638.69

shamsmosowi:
fixed on rc branch and .rowy.app

PR #584, opened 2021-11-17
fix csv import, items getting the same key

shamsmosowi:
#583


PHOENIX19824413 (approved):
Admit up , it all helps

Issue #563, opened 2021-10-15
Column menu update - Add accessible resize option


notsidney:
**Is your feature request related to a problem? Please describe.**
Columns can only be resized by drag on the edges, but this is not an accessible method and is especially difficult on touch screens.

It can also be difficult to see where the click/drag target exactly is. MUI’s Data Grid is a good example of explicitly showing the drag target: https://mui.com/components/data-grid/columns/#resizing

<img width="418" alt="Screen Shot 2021-10-15 at 17 54 50" src="https://user-images.githubusercontent.com/27017118/137444996-9b9f455a-a49b-467f-a437-199bf9a2f9ff.png">

**Describe the solution you'd like**
In the column menu, there should be an option to resize that opens a modal to set the column width, maybe with a preview.

![image](https://user-images.githubusercontent.com/47289133/190529762-f2368f2b-579a-479c-935d-ad375f628e0a.png)

gibsonliketheguitar:
I will claim this task

ashutosh887:
I would like to work on this Issue!

nithinrdy:
@harinij @BeeBombshell , mind if I take a stab at this? I've looked through the tanstack API, but there doesn't appear to be a method or something to directly update a column's width programmatically -- resizing via dragging seems to be the only supported way of doing this (using `getResizeHandler()`).

However, I've been able to hack together a way to update a column's width by constructing a fake MouseEvent with the desired value, and then dispatching said event to the handler I mentioned above. It works as expected.

If this approach looks good to you, I'd like to open a PR -- If possible, I'd also appreciate any designs to go off of :)
If not, I'd like your thoughts on what I could be missing here. Thanks!

nithinrdy:
I'm assuming actual code might help facilitate a discussion better than a simple comment, opened one.

PR #606, opened 2021-12-20
feat-563: Add option to manually resize column in column menu

gibsonliketheguitar:
The feature is pretty much done. #563
- Add an option called 'Resize column' in the column menu.

However, there is a bug with React Data Grid.
After dragging and resizing the column, you will NOT be able to manually update the column width with this feature.

I've confirmed there is no issue with rowy state manually and with react-dev tools and 
I've also built a [smaller version ](https://github.com/gibsonliketheguitar/demos/tree/master/rdg-test)of this feature to confirm the bug.
I started digging into the library, and believe React.memo is doing something weird. Will do the more extensive tests when I have time. 

I will open a [ticket](https://github.com/adazzle/react-data-grid/issues/2771) with React-Data-Grid and see if they can help.


notsidney:
Hi @gibsonliketheguitar, is our use of React.memo on the table preventing this re-render from occurring? Can we just add another dependency to the memo?

gibsonliketheguitar:
@notsidney its RDG's memo function handling column resize. I build a smaller application to confirm  the issue 

I can double check ROWY memos.

Issue #558, opened 2021-10-12
Add column to left/right is adding to the wrong place and is unpredictable


notsidney:
**Describe the bug**
When using the Add new [column] to left/right option in the column menu, it doesn’t always add to the right place and sometimes works.

gibsonliketheguitar:
Claiming this task

PR #610, opened 2021-12-27
fix-bug-558: adding column left or right unpredictable state

gibsonliketheguitar:
#558 
- applying the single responsibility principle, to improve maintenance in the future.  
- Introduce new table action called 'insert' to replace 'handleSave' in NewColumn Component

## Reason

handleModalSave  === tableActions.column.update === updateColumn

```ts
 const menuModalProps = {
   ...
    handleSave: handleModalSave,
  };

const handleModalSave = (key: string, update: Record<string, any>) => {
    actions.update(key, update);
};
 
const actions: TableActions = { 
...
column: {
      ...
      update: configActions.updateColumn,
    },
}

const updateColumn = (key: string, updates: any) => {
    const { columns } = tableConfigState;

    const updatedColumns = {
      ...columns,                                   // this maintains previous index
      [key]: { ...columns[key], ...updates },       // this allows for duplicate index
    };
  };
```

Bug
- a copy of a new column can result in duplicated indexes
-  after copy, the other column's index has to be updated 

Solution
- calculate the insert position, given left or right position,  and handle the edge case
- use _orderBy lodash to covert  columns into a data structure where we can use .splice on
- after insert, use reduce to rewrite index and return data into object format for dispatcher function


src/hooks/useTable/useTableConfig.ts
```diff
@@ -88,6 +89,8 @@ const useTableConfig = (tablePath?: string) => {
     column.width = width;
     let updatedColumns = columns;
     updatedColumns[column.key] = column;
+
+    console.log("what is updatedColumns resiz3", updatedColumns);
```
gibsonliketheguitar, line 93: remove console.log

src/hooks/useTable/useTableConfig.ts
```diff
@@ -112,6 +115,32 @@ const useTableConfig = (tablePath?: string) => {
       data: { columns: updatedColumns },
     });
   };
+
+  /** insert column by index
+   * @param col     properties of new column
+   * @param source  source object { index: selected index, position: left | right }
+   */
+
+  const insert = (col, source) => {
```
gibsonliketheguitar, line 124: double check col insert, with the last column header

src/hooks/useTable/useTableConfig.ts
```diff
@@ -112,6 +115,32 @@ const useTableConfig = (tablePath?: string) => {
...
+    const { columns } = tableConfigState;
+    const position = source.position === "right" ? 1 : 0;
+    const insertPos = source.index + position;
+    const targetPos = insertPos === -1 ? 0 : insertPos;
```
gibsonliketheguitar, line 128: document logic here

gibsonliketheguitar:
wait for #626 to be merge. 626 contains fix for keeping the index correct. 
This feature breaks when columns are delete because index are not updated

shamsmosowi (approved):

Issue #559, opened 2021-10-13
Reordering columns has unpredictable behavior


notsidney:
**Describe the bug**
When reordering columns using drag and drop, it sometimes doesn’t do anything and sometimes puts it in the wrong place, off by one. Performing the same reorder action corrects it, so users need to reorder the column twice for it to work.

gibsonliketheguitar:
Claiming this task

PR #611, opened 2021-12-28
fix-bug #559 unpredictable state of tableAction.reorder's function

gibsonliketheguitar:
#559 
- Use .reduce() to carry out data swap, and 
- transform columns into a correct data structure for dispatch. 

Details about Data swap
- During interaction, 
      if curr.key === a.key, then copy data[b],  index 
      vice versa === b.key, acc[a.key] =  { ...data[a], index }
  
I do request some clarification on the array mover function.
- When is there an undefined element?

I don't fully understand the arrayMover function. 
I think I am missing an edge case, assistance would be awesome.

### DIFF
--- a/src/hooks/useTable/useTableConfig.ts
+++ b/src/hooks/useTable/useTableConfig.ts
@@ -6,7 +6,7 @@ import _sortBy from "lodash/sortBy";
 
 import useDoc, { DocActions } from "../useDoc";
 import { FieldType } from "@src/constants/fields";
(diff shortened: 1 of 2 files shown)


notsidney (changes requested):
For reorder, we want to move a column to a new position and push all columns right to make space, not swap the two columns’ positions.

gibsonliketheguitar:
Will do

gibsonliketheguitar:
ready-im-ready

shamsmosowi (approved):

Issue #598, opened 2021-12-13
Cmd K Bar for Rowy


harinij:
**Is your feature request related to a problem? Please describe.**
It would great to have a cmd+K bar for Rowy to search and do most commonly used actions like create table, extensions. Filter, sort etc.

Exploration of different options for implementing this and discussion welcome.

gibsonliketheguitar:
I began a basic implementation of HotKeys. I notice that there is already a custom hook and Component for Hotkeys.

harinij:
This issue is still available for Good First issue. A library like this can be used for implementation https://github.com/timc1/kbar

Hardiegogo:
This is interesting, I would like to take this up.
Can you assign this to me? @harinij

Hardiegogo:
In this demo, I have implemented major navigation shortcuts throughout the app using kbar:
[Demo Link](https://github.com/rowyio/rowy/assets/51075671/979f18aa-7773-4733-8e5c-f7047b807a9e)

Apart from this, I am considering the following approach to implement specific actions on the pages using kbar:

- Creating a global atom called "hotkeys." This will be an object containing boolean state variables based on various actions.
- We can then use these properties in combination with useEffect to perform actions immediately upon opening specific pages when the user interacts with the kbar.
- These actions may include opening a pre-toggled view or perhaps opening the "createTable" modal directly through kbar.

I would appreciate your opinions and any feedback on this. @harinij  @Chadha93

PR #612, opened 2021-12-28
Feat#598:  implementation of HotKeys

gibsonliketheguitar:
#598 
Implement Hotkey as a Higher Order Component.  An alternative route would be using making a custom hook.

In addition, I tried hooking up addRow but it's being buggy. 
And in the future, we will have to decide 
1. create ref to call respective buttons/actions 
2. replicate the same ref style as AddColumn

I notice that there is already a useHotKey hook,  as well as components.
What's going on there?


shamsmosowi:
useHotKey was used for copy/cut/paste shortcut, I think it was disabled during a RDG migration due to a breaking change, needs to be reimplemented, feel free to give it a try

Issue #600, opened 2021-12-13
Right click menu for cells


harinij:
Add right click menu for table cells with useful features such as..

copy, paste
copy column name, id, ref 

other ideas welcome to discuss

gibsonliketheguitar:
Should we have a global state in Jotai to store the copying of values

shamsmosowi:
hi @gibsonliketheguitar, the copied values should be stored into the clipboard, so users are able to move their data out of rowy.
We should be keeping the same style as the column menu, but no need to include column level functionality included in the column menu already.

src200:
Opened a PR https://github.com/rowyio/rowy/pull/621

gibsonliketheguitar:
> Opened a PR #621

Was working on this to  #622. 
In addition, @src200 I notice you wrapped the Row with the ContextMenu.

The potential problem is the necessary duplication on all the <Row />
My implementation followed the ColumnMenu Code, and used  a CellMenuRef to handle the conditional rendering of the ContextMenu


@shamsmosowi  @src200 Let's figure out who will be taking this ticket

PR #623, opened 2022-01-06
feat-600 add context menu for cells

gibsonliketheguitar:
#600 
Implementation of Context Menu for Row's Cell

ContextMenu Component Wraps DataGrid's Row Component in Table Row Component.
Used a similar structure as the Column Menu, using a single CellMenuRef for the anchor position
Currently copied text is grabbed by calling anchorEl.innerHTML

Problems in the foreseeable feature
1. Need to think through if we want to integrate with #598 and other hotkey functions
2. Selecting the cell column.key is blocking me because the rowRendererProps passed to React Data Grid's Row does not contain column property. We need to figure out how to find the column.key. We can probably do a hacky function with DataGrid's onRow click, but I feel like there has to be a better.

### DIFF
--- a/src/assets/icons/Paste.tsx
+++ b/src/assets/icons/Paste.tsx
@@ -1,0 +1,10 @@
+import SvgIcon, { SvgIconProps } from "@mui/material/SvgIcon";
+import { mdiContentPaste } from "@mdi/js";
+
+export default function Paste(props: SvgIconProps) {
+  return (
+    <SvgIcon {...props}>
+      <path d={mdiContentPaste} />
+    </SvgIcon>
+  );
+}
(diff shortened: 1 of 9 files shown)


shamsmosowi:
Copying inner html is an interesting idea, but it wouldn't work well for complex cells images, color etc.
Have you had a look at the `onSelectedCellChange` RDG method? could be used to store reference of the cell in jotai state for the menu to use.

gibsonliketheguitar:
@shamsmosowi thanks for lead on, "onSelectedCellChange". It worked like a charm.

Currently implemented copy, paste, and paste from clipboard. We will need to think about which Icon to choose from paste and paste from the clipboard. 

A side note. I notice that I could not copy and paste with cmd+x & cmd+v. Can you check if this is occurring on your end too.

gibsonliketheguitar:
#630 New Branch, built on top of develop branch.

PR #630, opened 2022-02-01
Right click menu

gibsonliketheguitar:
#600 

Enable Right click on designated Field Types

PR #632, opened 2022-02-02
Right click menu -v2

gibsonliketheguitar:
- [Update conextMenuAction's type definition ](https://github.com/rowyio/rowy/pull/630#discussion_r797240195)
- [Add snackbar](https://github.com/rowyio/rowy/pull/630#discussion_r797239969)

- Convert to Jotai
   - [Comment 1](https://github.com/rowyio/rowy/pull/630#discussion_r797236871)
   - [Comment 2](https://github.com/rowyio/rowy/pull/630#discussion_r797274872)
- [Add optional shortcut label](https://github.com/rowyio/rowy/pull/630#discussion_r797235927)
- [Set MenuList to dense](https://github.com/rowyio/rowy/pull/630#discussion_r797234880)
- [Rename MenuRow Component to ContextMenuItem](https://github.com/rowyio/rowy/pull/630#discussion_r797234537)
- [Covert to async/await style](https://github.com/rowyio/rowy/pull/630#discussion_r797274381)


src/components/Table/ContextMenu/index.tsx
```diff
@@ -1,56 +1,28 @@
...
 
-  if (!contextMenuRef.current || hasNoActions) return null;
-
+  if (!contextMenu || hasNoActions) return <></>;
   return (
     <MenuContents
-      anchorEl={anchorEl}
-      open={open}
-      handleClose={handleClose}
+      anchorEl={anchorEl as HTMLElement}
```
notsidney, line 23: Side note: I think if you add `!anchorEl` as a condition on line 20, TypeScript won’t make you explicitly state `as HTMLElement` here.

Logically we know that if `contextMenu` is falsy, then `contextMenu.anchorEl` would not exist, but since you use `anchorEl` directly, TypeScript isn’t quite smart enough to know that they are the same.

src/components/Table/TableRow.tsx
```diff
@@ -1,18 +1,21 @@
-import { useProjectContext } from "@src/contexts/ProjectContext";
+import useContextMenuAtom from "@src/hooks/useContextMenuAtom";
 import { Fragment } from "react";
 import { Row, RowRendererProps } from "react-data-grid";
 
 import OutOfOrderIndicator from "./OutOfOrderIndicator";
 
 export default function TableRow(props: RowRendererProps<any>) {
-  const { contextMenuRef }: any = useProjectContext();
+  const { setContextMenu } = useContextMenuAtom();
```
notsidney, line 8: Same comment as `src/components/Table/index.tsx` regarding `useUpdateAtom`

src/components/fields/_BasicCell/BasicCellContextMenuActions.tsx
```diff
@@ -5,56 +5,63 @@ import CopyCells from "@src/assets/icons/CopyCells";
...
+import { SelectedCell } from "@src/atoms/ContextMenu";
 
-export default function BasicContextMenuActions() {
-  const { contextMenuRef, tableState, deleteCell, updateCell } =
-    useProjectContext();
+export interface IContextMenuActions {
+  label: string;
+  icon: React.ReactNode;
+  onClick: () => void;
+}
```
notsidney, line 15: Note for the future: when we have more context menu actions for different field types, we should move this to `src/components/fields/types.ts`. I’m also not sure about having this file be inside `_BasicCell`. This is fine for now however.

src/components/fields/_BasicCell/BasicCellContextMenuActions.tsx
```diff
@@ -5,56 +5,63 @@ import CopyCells from "@src/assets/icons/CopyCells";
...
+    } catch (error) {
+      enqueueSnackbar("Failed to copy", { variant: "error" });
+    }
     handleClose();
   };
 
-  const handleCut = () => {
-    const cell = selectedRow?.[selectedCol.key];
+  const handleCut = async () => {
     const notUndefined = Boolean(typeof cell !== "undefined");
```
notsidney, line 44:
```suggestion
    const notUndefined = cell !== undefined;
```

The `!==` operator should always return a boolean, so it’s not necessary to use the `Boolean` object for conversion here. You can also compare directly to `undefined` without using `typeof`

notsidney (changes requested):

src/components/Table/TableRow.tsx
```diff
@@ -1,25 +1,24 @@
...
 
 export default function TableRow(props: RowRendererProps<any>) {
-  const { contextMenuRef }: any = useProjectContext();
+  const { setAnchorEle } = useSetAnchorEle();
   const handleContextMenu = (
     e: React.MouseEvent<HTMLDivElement, MouseEvent>
   ) => {
     e.preventDefault();
-    if (contextMenuRef?.current) contextMenuRef?.current?.setAnchorEl(e.target);
+    setAnchorEle?.(e?.target as HTMLElement);
```
notsidney, line 13:
```suggestion
    setAnchorEle(e?.target as HTMLElement);
```

I thought `setAnchorEle` would always exist/never be `undefined`

src/atoms/ContextMenu.ts
```diff
@@ -0,0 +1,40 @@
...
+    await resetAnchorEle();
+    await resetSelectedCell();
+  };
+
+  return {
+    anchorEle,
+    selectedCell,
+    resetContextMenu,
+  };
+}
```
notsidney, line 40: I’m not sure of the benefits of creating custom hooks here, instead of accessing the atoms directly using `useUpdateAtom` in `Table` for example

notsidney (approved):

src/atoms/ContextMenu.ts
gibsonliketheguitar: I don't recall where I saw. But I pick up this pattern a while back because it's used to hide the imports 


```jsx
//no custom hook
import { anchorEleAtom } from '@src/store/contextmenu'
import { useAtom } from 'jotai'

const [anchorEle] = useAtom(anchorEleAtom)

```

//custom hooks
```jsx
import { useContextMenu } from '@src/store/contextmenu'

const { anchorEle } = useContextMenu()
```

it's purely syntactic sugar I think. We can remove it if necessary

src/components/Table/TableRow.tsx
gibsonliketheguitar: You are right. Ill change it

PR #640, opened 2022-02-07
Fix delete column, multi-options, and remove validation of duplicates

gibsonliketheguitar:


src/components/fields/Status/ConditionModal.tsx
```diff
@@ -19,14 +19,11 @@ export default function ConditionModal({
     setModal(EMPTY_STATE);
   };
   const handleAdd = () => {
-    const labelIsEmpty = Boolean(modal.condition.label.length < 4);
+    const labelIsEmpty = Boolean(modal.condition.label.length === 0);
```
gibsonliketheguitar, line 22: Should we remove this?

shamsmosowi (approved):

PR #654, opened 2022-02-17
Feat: Add Validation to Status Field's Conditional

gibsonliketheguitar:
Replace modal system with Rowy Form Builder
Add validation to 
- type of input data
- boolean
- string
- label

validation checks if there is duplicate labels, boolean, string values

TODO: validation needs to be added to number type.
In order to validate number field type, operator value and number value must be watch, in order to validate against existing value


src/components/fields/Status/utils/formatDataHelper.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * We have to format data value because,
+ * duplicate field types with same { name: [value]}  does something to validation
+ * we have to name values to booleanValue, stringValue, numberValue to get validation to work
```
gibsonliketheguitar, line 4: Update comments here

src/components/fields/Status/utils/validationHelper.ts
```diff
@@ -0,0 +1,76 @@
...
+    //cannot stringify sometime that does not exist
+    if (curr.value === undefined) result = { ...curr, value: "undefined" };
+    else result = { ...curr, value: JSON.stringify(curr.value) };
+    return result;
+  });
+
+  if (typeof index !== "number")
+    result = !_find(formatConditions, { value: value });
+  else {
+    //const copyConditions = [...formatConditions];
```
gibsonliketheguitar, line 71: remove unused code

PR #659, opened 2022-02-28
Fix initial values for nested fields

n-sviridenko:
We were unable to update values of select fields pointing to nested keys and lost a lot of data as the UI was showing that it's saved, but in fact it wasn't.


n-sviridenko:
@harinij @notsidney @shamsmosowi we might be not the only ones losing their data.

notsidney (approved):

notsidney:
Thanks for this. I’ve released it to rowy.app and it will be part of a v2.4.0 release by the end of the week.

Issue #519, opened 2021-09-30
Slider Field - Add customization option


harinij:
**Is your feature request related to a problem? Please describe.**
Currently, the "Slider" field does have customization to change the color from red for 0% and green for 100% to anything else
<img width="181" alt="Screen Shot 2021-10-01 at 2 24 33 am" src="https://user-images.githubusercontent.com/307298/135494022-679f2c44-b24a-497c-833d-6488e8ace906.png">

**Describe the solution you'd like**
Ability to pick color from starting and ending percentage and mid percentage. And the color for the in between range will be auto calculated. SImilar to the implementation on the percentage fiel.d

ashutosh887:
Hey @htuerker @harinij I would like to contribute! 😃

htuerker:
Hey, @ashutosh887! Great to hear that! Please, do not hesitate to reach out to me for anything you'd like to ask.

ashutosh887:
How do I set it up locally on my system for Development purposes? @htuerker

harinij:
@ashutosh887 you can follow the instructions for manual install [here](https://docs.rowy.io/setup/install#option-2-manual-install)

harinij:
Hi @ashutosh887 removing for now due to lack of activity.

ashutosh887:
I would like to work on this again
Actually couldn't get help from the community regarding some of my doubts
@harinij @htuerker

harinij:
Hi @ashutosh887 I have reassigned for now. In case of any questions, you can ask on the #contribution channel on Rowy Discord rowy.io/discord

ashutosh887:
Thanks @harinij

mnmt7:
@ashutosh887, are you still working on this? If not, @harinij, can I have the issue assigned to me?

ashutosh887:
Yes @mnmt7 
Creating PR soon

Issue #517, opened 2021-09-30
Toggle field - Add customization options


harinij:
**Is your feature request related to a problem? Please describe.**
Currently, the "Toggle" field doesn't have customization to change the color from green to anything else

**Describe the solution you'd like**
Allow in the UI option for users to add color using the color pick modal similar to what is used in the percentage field
![Screenshot 2023-04-20 at 7 07 44 pm](https://user-images.githubusercontent.com/307298/233317767-90a4131d-39cd-421f-a749-2de92f309fa5.png)

yasinbordbar:
Please assign me to do this.

ChinazoOnwukaike:
Can this be assigned to me please?

harinij:
@ChinazoOnwukaike please let us know if you are still working on this or if you need any inputs. Unassigning for now.

RajGM:
Hi @harinij and @notsidney 
The toggle field doesn't have color customization in the column config.

RajGM:
Hi @harinij 
I am working on this: https://github.com/rowyio/rowy/commit/ef4fa7d4818d118f9be304033847096904d9217c

Color pick modal is added.
Remaining: Coloring of the whole field based on true/false;
![ColorModal](https://user-images.githubusercontent.com/16620150/233811318-42eec3ff-7c09-4f68-94f0-33d16177ee21.jpg)

RajGM:
Hi @harinij 

`Is your feature request related to a problem? Please describe.
Currently, the "Toggle" field doesn't have customization to change the color from green to anything else`

Does that means we are only looking to change the color or toggle button instead of whole cell?

RajGM:
![Switch](https://user-images.githubusercontent.com/16620150/233948937-3a2e6ebe-4c47-4931-ac72-5e74e2ff51af.jpg)

RajGM:
Due to the defined type in material UI, we are supposed to pass only: 

(property) SwitchProps.color?: OverridableStringUnion<"primary" | "secondary" | "error" | "info" | "success" | "warning" | "default", SwitchPropsColorOverrides> | undefined

RajGM:
Hi @shamsmosowi
I am trying to change the color but it override the whole thing.

const theme = useTheme();
  const { colors } = (column as any).config;
  const percentage = typeof value === "number" ? value : 0;
  const finalColor = resultBoolColorsScale(
    percentage,
    colors,
    theme.palette.background.paper
  ).toHex();
[...]

PR #692, opened 2022-06-22
Feat: Percentage field color customization

### LINKED ISSUE #518: Percentage Field - Add customization option
Currently, the "Percentage" field does have customization to change the color from red for 0% and green for 100% to anything else
Screen Shot 2021-10-01 at 2 22 33 am

Ability to pick color from starting and ending percentage and mid percentage. And the color for the in between range will be auto calculated.

https://github.com/rowyio/rowy/tree/develop/src/components/fields/Percentage

htuerker:
Hello! 👋 

This fixes #518. 

Changes I made in order:
1. First, to meet the config requirement, I converted the `Percentage` field from `BasicCell` to `TableCell`.
2. I stored color configs like below:
```javascript
config: {
   startColor: { hex, hsl, rgb } as Color;
   midColor: { hex, hsl, rgb } as Color;
   endColor: { hex, hsl, rgb } as Color;
}
```
3. I almost entirely copied the color picker logic from `fields/Color/SideDrawerField` 
4. I created `Settings` and add it to the fields config as required.
5. I also extracted the ColorPickerInput for future usage i.e #517 #519. As I copied it from `Color` field, maybe we can try to use it there too.
6. There's no broken part against the new `resultColorsScale` utility logic as I handled this by giving a default `colors` parameter. Everything works fine. 
7. I fixed `emotion-cache` problems by debouncing the dynamic style changes with [8854965](https://github.com/rowyio/rowy/pull/692/commits/8854965109b2910d1413f5ccc08562fbf4418488) 
8. I added width responsiveness to `ColorPicker` container with [fce32f4](https://github.com/rowyio/rowy/pull/692/commits/fce32f4a49a670336fdaf334a8800a285aa7ddff)

I'm pretty much new to the codebase. I'd love to get a little more informative review. I'll be able to work on possible change requests. 

Change previews: 
Screen Shot 2022-06-22 at 23 19 50
Screen Shot 2022-06-23 at 17 41 03
Screen Shot 2022-06-22 at 23 21 50


src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,70 @@
...
+    ...config,
+  };
+
+  const [start, setStart] = useState<Color>(startColor);
+  const [mid, setMid] = useState<Color>(midColor);
+  const [end, setEnd] = useState<Color>(endColor);
+  return (
+    <>
+      <InputLabel>
+        Start Color
```
htuerker, line 34: I'm pretty bad at naming things. I'd love to change these labels with a more meaningful alternative.

tinacai97:
Hi @htuerker! I'm Tina, the designer here at Rowy. 
We're so glad to see you contributing to our project, thank you!

From a design standpoint, I have a couple of minor comments. 

1. The color picker "appears" to be cut off. I don't think it actually is, but there seems to be some sort of illusion of cut-off because the bottom of the drop down has a rounded corner, while the top appears to stem straight from the drop down bar. I wonder if we could distance the color picker from the drop down and add some padding and a rounded corner... or whatever workaround you can think of!

Percentage_Comment_1

2. I wonder if we can give users "optional" color sections. Perhaps I'd like two toned percentages, or maybe even one. It'd be cool to be able to have an easier option for that, rather than just setting the section I'd like "off" to fully transparent. Thought of checkboxes on the top of my head, but maybe you've got a better solution out there. 

Percentage_Comment_2

Thanks again for contributing to Rowy! We're an open-source community because of the awesome things people like you are able to add 😄

htuerker:
> The color picker "appears" to be cut off.

I must admit that @tinacai97 you have sharp eyes(.33px) 🤩 I think `react-color-palette`'s calculations cause this precision mistake. I put `min-width: 100%` and `max-width: 100%` to fix this, and it seems solved. 
Screen Shot 2022-06-24 at 01 24 43

Also, here's how it looks with top margin, let me know @tinacai97 which one fits well.
Screen Shot 2022-06-24 at 01 29 22

There's still an illusion on the color palette's both left and right borders. The only solution I can bring here is giving padding:
Screen Shot 2022-06-24 at 01 46 51


> I wonder if we can give users "optional" color sections. Perhaps I'd like two-toned percentages, or maybe even one. 

I liked that idea. So, as default, we can simply set 3-colors(red-yellow-green). Then, leave it to the user to pick by checkboxes if wanted. Selecting from checkboxes would make it so easy to set single or double-toned variations. 
Screen Shot 2022-06-24 at 01 56 04
Screen Shot 2022-06-24 at 02 01 31

Basic logic: 
1. If there's only one tone selected i.e `red`: create a tone from `white`(works fine with dark mode as well) to `red`(success color) based on the percentage value
2. If there are two selected i.e `red` and `blue`: derive mid color from them, then create tone from `red` to `derived color` to `blue`
3. If there're three selected, no need to change the current logic. 

Screen Shot 2022-06-24 at 00 04 00


Thanks for your suggestions as well as a warm welcome! @tinacai97 ❤️

tinacai97:
Hi @htuerker!

> There's still an illusion on the color palette's both left and right borders. The only solution I can bring here is giving padding: Screen Shot 2022-06-24 at 01 46 51

🥳 This is much clear-er! I definitely like the version with the padding. 

My one tid-bit is if it would make more sense to keep the right angles at each corner of the actual color picker. I ask because aaaallll the way to the bottom-left would be black, #000, but would aaaallll the way to the top-left be white, #fff? I would imagine it might be slightly off white with the rounded corner there, though the hex code would tell me if that was the right assumption. But that also makes me wonder if there are some colors missing in those corners 🤔 

Otherwise though, the color picker  being encapsulated in the same rounded corner container is awesome!

Percentage_Comment_3

> > I wonder if we can give users "optional" color sections. Perhaps I'd like two-toned percentages, or maybe even one.
> 
> I liked that idea. So, as default, we can simply set 3-colors(red-yellow-green). Then, leave it to the user to pick by checkboxes if wanted. Selecting from checkboxes would make it so easy to set single or double-toned variations. Screen Shot 2022-06-24 at 01 56 04 Screen Shot 2022-06-24 at 02 01 31

💯 Perfect! 
 
> Thanks for your suggestions as well as a warm welcome! @tinacai97 ❤️

Of course, thank you for contributing! We hope you learned lots and got as much out of it as we did. I'll discuss pulling this with the team soon. Looking forward to any other contributions you make!

src/components/ColorPickerInput.tsx
```diff
@@ -0,0 +1,92 @@
...
+        fieldSx,
+        {
+          marginTop: "1rem",
+          padding: "1rem",
+          borderColor: "divider",
+          transitionDuration: 0,
+          "& .rcp": {
+            border: "none",
+            "& .rcp-saturation": {
+              borderRadius: "4px",
```
htuerker, line 75: It'd be better to not use absolute units

src/components/ColorPickerInput.tsx
```diff
@@ -0,0 +1,92 @@
...
+            },
+            "& .rcp-body": {
+              boxSizing: "unset",
```
htuerker, line 78: useless styling

src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,195 @@
...
+  children,
+}: {
+  colorKey: string;
+  color: Color;
+  active: boolean;
+  setActive: (activeKey: string | null) => void;
+  disabled?: boolean;
+  children?: ReactElement;
+}) => {
+  const toggleCollapse = () => !disabled && setActive(active ? null : colorKey);
```
htuerker, line 35: Better to lift this logic to settings, so there'll be no need to pass active, setActive, or color key. Send only an onClickHandler instead.

src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,195 @@
...
+    if (disabled && active) {
+      setActive(null);
+      return false;
+    }
+    return active;
+  })();
+
+  return (
+    <Box sx={{ width: "100%" }}>
+      <ButtonBase
```
htuerker, line 47: A huge wrapper, better to extract it as a styled component

src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,195 @@
...
+              }),
+            "&.Mui-disabled": { color: "text.disabled" },
+          },
+          active && {
+            transitionDelay: "0s",
+            transitionDuration: "0s",
+          },
+        ]}
+      >
+        <Box
```
htuerker, line 70: Again extract the wrapper, it isn't lovely and makes the logic seem more complex. Better to stay away from composition.

src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,195 @@
...
+            transform: active ? "rotate(180deg)" : "none",
+          }}
+        />
+      </ButtonBase>
+      <Collapse in={toggleState}>{children}</Collapse>
+    </Box>
+  );
+};
+
+const defaultColors: { [key: string]: Color } = {
```
htuerker, line 95: Rename: configDefaults

src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  return (
+    <>
+      {Object.keys(checkStates).map((colorKey) => {
+        const color = config[colorKey] || defaultColors[colorKey];
+        return (
+          <Box key={colorKey} sx={{ display: "flex", flexDirection: "column" }}>
+            {colorLabels[colorKey]}
+            <Box
+              sx={{
```
htuerker, line 134: Inline stylings reduce readability here, since this is the most complex part better to use a style wrapper

src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,195 @@
...
+        );
+      })}
+      <InputLabel>
+        Preview:
+        <Box sx={{ display: "flex", textAlign: "center" }}>
+          {[0, 0.125, 0.25, 0.375, 0.5, 0.625, 0.75, 0.875, 1].map((value) => {
+            const { startColor, midColor, endColor } = config;
+            return (
+              <Box
+                key={value}
```
htuerker, line 174: Extract wrapper

htuerker:
Hello, again @tinacai97! 👋 I found some time to work on changes. There're a couple of bugs/misbehaviors right now. I'd love to hear your further comments on this:

https://user-images.githubusercontent.com/46192266/175763161-3f31ae5a-170d-44d9-8726-6529d5ada1e1.mov

There's only one functionality I'd like to add here which is the single-double tone selections without middle color.

src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,195 @@
...
+          }}
+        />
+      </ButtonBase>
+      <Collapse in={toggleState}>{children}</Collapse>
+    </Box>
+  );
+};
+
+const defaultColors: { [key: string]: Color } = {
+  default: toColor("hex", "#FFFFFF"),
```
shamsmosowi, line 96: This should be `#FFF0` for dark theme compatibility

shamsmosowi:
this is looking pretty cool! , instead of start/end colors, what about min/max colors?

src/components/ColorPickerInput.tsx
notsidney: This is in MUI’s [`sx` prop](https://mui.com/material-ui/react-box/#the-sx-prop), so you can access our theme’s border-radius property with `borderRadius: 1`, which will set it to `1 * 4px`. You can find [all the properties here](https://mui.com/system/properties/).

notsidney:
Hi @htuerker, this is a cracking PR! I work primarily on Rowy’s frontend and was the one who implemented the side drawer color picker you’re extending.

---

The only concern I have with this implementation is that you set the data structure for this config to be fixed to three items, start, mid, and end. I think it would be better to store it as an array of values instead, so we’re not constrained to only ever supporting 3 colors at most.

Then, the UI could be expanded in the future to support multiple color stops like a gradient editor in a design tool. Although the stops would be equidistant for simplicity.

Screenshot of Figma gradient editor

Supporting more than 3 colors would require the `resultColorsScale` util to be rewritten to support an arbitrary number of colors, so we should hold off on allowing the user to set more than 3. We can update this config UI at any time, but if we stick with the current data structure, we’d need to “migrate” this config to the new one.

Additionally, if you store a CSS-compatible color in the config UI, like a hex or rgba (if supporting alpha) string, then you don’t need to import colord to convert it to a hex string in the table cell components.

---

For the config UI itself, I would suggest two changes:

1. Keep the text that displays the color’s hex code to be the standard text color (black in light mode, white in dark mode). I think you just need to unset the `color` CSS property. In the video you sent, when the color was set to yellow, the text would fail accessibility requirements for color contrast. The swatch on the left is enough to show the color selected.
2. If possible, it might be good to display the three color pickers side by side and make it responsive to save space. Those pickers feel a bit too wide. You can look at [MUI’s `Grid` component](https://mui.com/material-ui/react-grid/) for this. (Note: this is used to lay out using a 12-column grid, which predates CSS Grid. This component actually uses CSS Flex) 

---

> to meet the config requirement, I converted the Percentage field from BasicCell to TableCell

You can keep the BasicCell component, but remove the coloring for it, so that a value still appears on scroll, instead of displaying `BasicCellNull`.

---

And just a side note on the design for the color picker: it was meant to appear as one large rounded rectangle when opened, which is why the top corners of the react-color-palette UI are not rounded. This also removes the issue with parts of the color canvas being cut off by the rounded corners.

It appears that the styles are being overridden somewhere, or I could have copied the wrong styles, because the bottom corner radii don’t match the collapsed box (`theme.shape.borderRadius * 1`). Also, dark mode styles don’t seem to be working :/ This is what it’s meant to look like:

Screenshot of corrected color picker

Ideally, it would closely match the design of other dropdown pickers like single and multi select, but those use a [`Popover`](https://mui.com/material-ui/react-popover/). I think I found it easier to implement it as a collapsible element at the time.

Screenshot of select design

That may also be a workaround to the issue with dynamic resizing—just set it to a fixed size all the time, like with the current date pickers (the MUI component isn’t easily resizable). The side drawer color picker is fixed at 440px, I believe.

Screenshot of date picker design

---

I quite like the approach you took to bring out `ColorPickerInput` into its own component. That’s definitely the best way to do this, especially as you noted it will come in handy for future PRs to the other issues. I’m also quite impressed that you were able to understand our codebase, which at the moment isn’t very well documented!

tinacai97:
Hi @htuerker, 

Those changes look great! Except I might have not been clear with the color palette corners. I was trying to show in the image that we should keep them at right angles as to prevent any missing colors. See below:

Screenshot 2022-06-27 104933

Otherwise, it _looks_ awesome to me! @notsidney can speak for the bugs and misbehaviors much better than I can.

tinacai97:
> And just a side note on the design for the color picker: it was meant to appear as one large rounded rectangle when opened, which is why the top corners of the react-color-palette UI are not rounded. This also removes the issue with parts of the color canvas being cut off by the rounded corners.
> 
> It appears that the styles are being overridden somewhere, or I could have copied the wrong styles, because the bottom corner radii don’t match the collapsed box (`theme.shape.borderRadius * 1`). Also, dark mode styles don’t seem to be working :/ This is what it’s meant to look like:
> 
> Screenshot of corrected color picker

Hi @notsidney,

Just wanted to add here that the suggestion to separate the color picker from being directly adjacent to the drop down was by me 😅. The color picker felt a bit "cut off" when positioned there in my perspective. Additionally the actual circle indicator overlaps with some of the drop down when choosing a color at the top of the palette, as in your screenshot. So that is probably why some of the style changed and the color picker ended up getting some rounded corners. 

Just wanted to add my two cents, I'll let you take it from here. I'm sure you know what is better design consistency wise!

htuerker:
> Those changes look great! Except I might have not been clear with the color palette corners. I was trying to show in the image that we should keep them at right angles as to prevent any missing colors. See below:

It's good now. There was exactly a problem at that time you asked.

I'm preparing a PR for the library itself.

https://user-images.githubusercontent.com/46192266/176062317-5d6b5cd6-a5c8-4b05-849e-f38c945e4e4d.mov

htuerker:
@notsidney I guess this is fine now! What a good first issue 😅

Last improvements:
1. Add BasicCell without background color on scroll instead BasicCellNull
2. Replace Collapse with TextField select
3. Add grid layout for picker checkboxes(md-4 sx-12)
4. Still favor a responsive color picker as a non-responsive one makes UI pretty ugly on the Color field's sidebar
5. For now, the overall state in Settings is hard-coded as we have the exact 3 inputs. But we can now easily expand it to support more colors.

src/components/fields/Percentage/SideDrawerField.tsx
```diff
@@ -34,11 +35,14 @@ export default function Percentage({
                 width: 20,
                 height: 20,
                 borderRadius: 0.5,
-                boxShadow: (theme) =>
-                  `0 0 0 1px ${theme.palette.divider} inest`,
+                boxShadow: `0 0 0 1px ${theme.palette.divider} inest`,
                 backgroundColor:
                   typeof value === "number"
-                    ? resultColorsScale(value).toHex() + "!important"
+                    ? resultColorsScale(
```
notsidney, line 41: Just note that in the MUI `sx` prop, you can pass in a function with `theme` as the first argument to access the theme object for any CSS property, so you didn’t need to change the `boxShadow` definition here or use `useTheme`

src/components/fields/Percentage/BasicCell.tsx
```diff
@@ -1,41 +1,23 @@
...
+        style={{
+          textAlign: "right",
+          color: theme.palette.text.primary,
+          position: "relative",
+          zIndex: 1,
+        }}
+      >
+        {Math.round(percentage * 100)}%
+      </div>
+    </>
```
notsidney, line 21:
```suggestion
      <div
        style={{
          textAlign: "right",
          color: theme.palette.text.primary,
          position: "relative",
          zIndex: 1,
        }}
      >
        {Math.round(percentage * 100)}%
      </div>
```

You don’t need to wrap this in a Fragment

src/components/ColorPickerInput.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const [localValue, setLocalValue] = useState(value);
+  const [width, setRef] = useResponsiveWidth();
+  const theme = useTheme();
+  const debouncedOnChangeComplete = useDebouncedCallback((color) => {
+    handleOnChangeComplete(color);
+  }, 100);
+
+  useEffect(() => {
+    debouncedOnChangeComplete(localValue);
+  }, [debouncedOnChangeComplete, localValue]);
```
notsidney, line 59: I’m not sure why this is done this way. react-color-picker already has a [`onChangeComplete` prop](https://github.com/Wondermarin/react-color-palette#colorpicker-props) fired on mouseup. Then you can avoid debouncing and using an effect.

If you need to debounce (when writing to db), it should be handled in the parent component. If the reason for debounce is to improve performance, you can use React 18’s [`startTransition`](https://reactjs.org/docs/react-api.html#starttransition)

```suggestion
```

src/components/ColorPickerInput.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export interface IColorPickerProps {
+  value: Color;
+  handleOnChangeComplete: (color: Color) => void;
+  disabled?: boolean;
+}
+
+export default function ColorPickerInput({
+  value,
+  handleOnChangeComplete,
```
notsidney, line 47:
```suggestion
  onChangeComplete: ColorPickerProps['onChangeComplete'];
  disabled?: boolean;
}

export default function ColorPickerInput({
  value,
  onChangeComplete,
```

Pass the `onChangeComplete` prop to the `<ColorPicker>` component directly. You also need to import `ColorPickerProps` from react-color-palette.

src/components/ColorPickerInput.tsx
```diff
@@ -0,0 +1,88 @@
...
+            },
+          },
+        },
+      ]}
+    >
+      <ColorPicker
+        width={width}
+        height={150}
+        color={localValue}
+        onChange={(color) => setLocalValue(color)}
```
notsidney, line 84:
```suggestion
        onChange={(color) => setLocalValue(color)}
        onChangeComplete={onChangeComplete}
```

src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,186 @@
...
+                  inset: 0,
+                  backgroundColor: resultColorsScale(
+                    value,
+                    colors,
+                    theme.palette.background.paper
+                  ).toHex(),
+                  opacity: 0.5,
+                }}
+              />
+              <Typography>{Math.floor(value * 100)}%</Typography>
```
notsidney, line 179:
```suggestion
              <Typography style={{ position: "relative", zIndex: 1 }}>
                {Math.floor(value * 100)}%
              </Typography>
```

It looks like the issue here with the text looking too light is because this appears below the colored Box in the z-axis. You can fix that by setting a z-index.

src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,186 @@
...
+    setCheckStates(
+      checkStates.map((value, idx) => (index === idx ? checked : value))
+    );
+  };
+
+  const handleColorChange = (index: number, color: Color): void => {
+    setColorsDraft(
+      colorsDraft.map((value, idx) => (index === idx ? color.hex : value))
+    );
+  };
```
notsidney, line 58: We’re getting into the weeds of React here—I don’t think you need to make a `colorsDraft` state, since that state definition only ensures that it’s not undefined and provides a default value.

You then sync up `colorsDraft` with the parent component using an Effect that calls `onChange`. This also isn’t necessary—you can read more about that in the [React docs](https://beta.reactjs.org/learn/you-might-not-need-an-effect#notifying-parent-components-about-state-changes). Disabling this eslint rule is also a code smell: `// eslint-disable-next-line react-hooks/exhaustive-deps`.

Instead, you could ensure the `colorsDraft` is a correct value without state:
```ts
const colors = config.colors ?? defaultColors;
```

Then, call `onChange` directly in your change handlers, so you’re only ever updating the state in the parent component [`ColumnConfig`](https://github.com/rowyio/rowy/blob/0b81f80e6cc16c324f92b2b2ceec3e7c72d227b2/src/components/ColumnModals/ColumnConfigModal/ColumnConfig.tsx#L43).

```suggestion
  const colors = config.colors ?? defaultColors;

  // It makes sense to keep this a state
  const [checkStates, setCheckStates] = useState<boolean[]>([
    colorsDraft[0],
    colorsDraft[1],
    colorsDraft[2],
  ]);

  const onCheckboxChange = (index: number, checked: boolean) => {
    onChange("colors")(
      colorsDraft.map((value: any, idx: number) =>
        index === idx ? (checked ? value || defaultColors[idx] : null) : value
      )
    );
    setCheckStates(
      checkStates.map((value, idx) => (index === idx ? checked : value))
    );
  };

  const handleColorChange = (index: number, color: Color): void => {
    onChange("colors")(
      colorsDraft.map((value, idx) => (index === idx ? color.hex : value))
    );
  };
```

src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  return (
+    <>
+      <GlobalStyles
+        styles={{
+          "& .MuiList-root.MuiMenu-list": {
+            padding: 0,
+          },
+        }}
+      />
```
notsidney, line 68: This shouldn’t be a part of GlobalStyles. It may affect other dropdowns. You can include it in the `<TextField>` `sx` prop below.

```suggestion
```

src/components/fields/Percentage/Settings.tsx
```diff
@@ -0,0 +1,186 @@
...
+                  },
+                ]}
+                onChange={() => onCheckboxChange(index, !checked)}
+              />
+              <TextField
+                select
+                label={colorLabels[index]}
+                value={1}
+                fullWidth
+                disabled={!checkStates[index]}
```
notsidney, line 103:
```suggestion
                disabled={!checkStates[index]}
                sx={{
                  "& .MuiList-root.MuiMenu-list": {
                    padding: 0,
                  },
                }}
```

Moved the styles out of GlobalStyles above

notsidney (changes requested):

src/components/fields/Percentage/SideDrawerField.tsx
htuerker: I did this change because I used the theme variable for the backgroundColor as well.

src/components/fields/Percentage/Settings.tsx
htuerker: This acts as a modal and is opened in the presentation div in the root. 

It has 4px padding on my screen. ColorPicker has 12px(spacing(1.5)). So, paddingTop(1) solves this in my practice. I'm just not sure about if MuiList gets as same as spacing(0.5) which is 4px here, right?

src/components/fields/Percentage/Settings.tsx
htuerker: Derived states just bring simplicity... 💯

htuerker: I've said to you that I read this article twice. What a shame 👀

src/components/ColorPickerInput.tsx
htuerker: The problem here is every color changes frame(100 times+ just for a little drag) appends lots of styles into head as we're generating styles regarding. I'm aware of this problem for now. This also makes hard to reuse this component I already tried it for Color field. 

I'm actually planning to work on another related issue after this. I've decided to think through this reusability problems during slider customization implementation. I'll research how we can use startTransition btw.

Also, a similar implementation used in Color field's PopoverCell.

src/components/ColorPickerInput.tsx
htuerker: This is a nice trick I did not know it, thanks!

htuerker: I think it's better not to use this as onChangeComplete is optional in ColorPickerProps.

src/components/fields/Percentage/SideDrawerField.tsx
notsidney: Yes, I meant you could do the same thing for `backgroundColor` as well: `backgroundColor: (theme) => ...`

src/components/ColorPickerInput.tsx
notsidney: > I think it's better not to use this as onChangeComplete is optional in ColorPickerProps.

You can use `NonNullable<ColorPickerProps['onChangeComplete']>` [(TypeScript docs)](https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype)

src/components/ColorPickerInput.tsx
notsidney: > The problem here is every color changes frame(100 times+ just for a little drag)

It only re-renders this ColorPickerInput component, which is expected and necessary. This is why we use `onChangeComplete` to save the selected color only when the user has stopped interacting with the picker, and then update the other components like preview.

> appends lots of styles into head as we're generating styles regarding.

This only happens when the user stops dragging (`onChangeComplete`), and MUI only appends styles to head in development mode. Notice that in dev mode, there are 1000+ nodes of `<style data-emotion="mui" data-s>` but these do not appear in production (like demo.rowy.io). So this isn’t a problem.

> Also, a similar implementation used in Color field's PopoverCell.

This is because when `onChangeComplete` is called there, it immediately writes to the database, and we need to debounce writes for that since the user pays for writes. In this case, it’s all local, so we don’t have to add a debounce.

---

Additionally, now that you’ve accepted the suggestion to pass `onChangeComplete` directly, you’ve made this debounce + effect redundant. Adding logs, I saw that:
1. It fired when the dropdown is first opened.
2. It fired when the user stopped moving the mouse, but is still holding down the mouse button.
3. It fired when the user stopped holding the mouse button (this is the only time we want to fire), but since we already have an `onChangeComplete` prop, `onChangeComplete` is called twice at this point.

I’ve added another change request and when you click accept on that in GitHub, I’ll merge this PR.

src/components/ColorPickerInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const [width, setRef] = useResponsiveWidth();
+  const theme = useTheme();
+
+  const debouncedOnChangeComplete = useDebouncedCallback((color) => {
+    onChangeComplete(color);
+  }, 100);
+
+  useEffect(() => {
+    debouncedOnChangeComplete(localValue);
+  }, [debouncedOnChangeComplete, localValue]);
```
notsidney, line 60: This is made redundant by passing `onChangeComplete` directly. See discussion here: https://github.com/rowyio/rowy/pull/692#discussion_r912590033

```suggestion
```

notsidney (changes requested):
This should be the last change required before I can merge this PR.

PR #696, opened 2022-06-28
fix(rich-text-editor): fix dark mode ui appearance

### LINKED ISSUE #537: Bug in RichText field Headers option
Currently, when someone tries to edit a RichText field in the side drawer, the headers drop down list has a CSS color issue where it is all white in light and dark mode. Needs to fixed

Screen Shot 2021-10-01 at 11 37 35 am

htuerker:
Fixes #537

https://user-images.githubusercontent.com/46192266/176060365-763e675e-df95-44d4-8d97-05e4a6fb7bc1.mov

This is a very minimal solution. I tried something else but it seemed overkill then left it like that.

@notsidney This gives at least a chance to make a theme switch without issues.

For future references on this issue, there might be a solution something like the below:

```javascript
const useDisabledStylesheets(mode, sheet, cache) {
  useEffect(()=> {
    if(!cache[mode]) {
        cache[mode] = sheet;
    }
    sheet.disabled = false;
     return () => {
      cache[sheet].disabled = true;
    }
  })
}

import "/foo.css"
const LightCSS = (cache) => {
   useDisabledStylesheets("light", document.stylesheets[document.stylesheets.length - 1]), cache]
   return null;
}


```

### DIFF
--- a/src/components/RichTextEditor.tsx
+++ b/src/components/RichTextEditor.tsx
@@ -121,6 +121,40 @@ export default function RichTextEditor({
   return (
     <Styles focus={focus} disabled={disabled}>
     <style>{theme.palette.mode === "dark" ? skinDarkCss : skinCss}</style>
+        <GlobalStyles
+          styles={{
+            ".tox": {
+              "& .tox-menu.tox-menu, &.tox-tinymce-aux .tox-toolbar__overflow.tox-toolbar__overflow":
+                {
+                  backgroundColor: theme.palette.background.paper,
+                  backgroundImage:
+                    "linear-gradient(rgba(255, 255, 255, 0.12), rgba(255, 255, 255, 0.12))", // elevation 8
+                  boxShadow: theme.shadows[8],
+                  border: "none",
+                  borderRadius: (theme.shape.borderRadius as number) * 2,
+                },
+
+              "& .tox-collection--list": {
+                "& .tox-collection__group.tox-collection__group": {
+                  padding: theme.spacing(0.5),
+                },
+                "& .tox-collection__item": {
+                  borderRadius: theme.shape.borderRadius,
+                },
+              },
+
+              "&.tox-tinymce-aux .tox-toolbar__overflow.tox-toolbar__overflow":
+                {
+                  padding: theme.spacing(0.5, 0),
+                },
+              "& .tox-tbtn.tox-tbtn": {
+                borderRadius: theme.shape.borderRadius,
+                margin: 0,
+              },
+            },
+          }}
+        />
+        
         <Editor
           disabled={disabled}
           init={{
--- a/src/components/RichTextEditor.tsx
+++ b/src/components/RichTextEditor.tsx
@@ -203,7 +203,6 @@ export default function RichTextEditor({
             if (onBlur) onBlur();
           }}
         />
-      </Suspense>
     </Styles>
   );
 }
--- a/src/components/RichTextEditor.tsx
+++ b/src/components/RichTextEditor.tsx
@@ -59,21 +59,25 @@ const Styles = styled("div", {
 
     "& .tox-toolbar__group": {
       border: "none !important",
+      "& .tox-tbtn": {
+        "&:hover:": {
+          backgroundColor: "inherit",
+        },
+        "&:focus": {
+          backgroundColor: "inherit",
+        },
+      },
       "& .tox-tbtn__select-chevron": {
         transition: theme.transitions.create("transform", {
           duration: theme.transitions.duration.short,
         }),
       },
       "& .tox-tbtn--select": {
-        "&::hover, &::focus": {
-          backgroundColor: theme.palette.action.input,
-        },
         "& .tox-tbtn__select-chevron": {
           transform: "none",
         },
       },
       "& .tox-tbtn--active": {
-        backgroundColor: theme.palette.action.input,
         "& .tox-tbtn__select-chevron": {
           transform: "rotate(180deg)",
         },
--- a/src/components/RichTextEditor.tsx
+++ b/src/components/RichTextEditor.tsx
@@ -156,6 +156,7 @@ export default function RichTextEditor({
         />
         
         <Editor
+          key={theme.palette.mode}
           disabled={disabled}
           init={{
             skin: false,
--- a/src/components/RichTextEditor.tsx
+++ b/src/components/RichTextEditor.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { lazy, Suspense, useState } from "react";
 
 import { styled, useTheme } from "@mui/material";
 import { Editor } from "@tinymce/tinymce-react";
@@ -9,8 +9,6 @@ import "tinymce/tinymce.min.js";
 import "tinymce/themes/silver";
 // Toolbar icons
 import "tinymce/icons/default";
-// Editor styles
-import "tinymce/skins/ui/oxide/skin.min.css";
 // Content styles, including inline UI like fake cursors
 /* eslint import/no-webpack-loader-syntax: off */
 import contentCss from "!!raw-loader!tinymce/skins/content/default/content.min.css";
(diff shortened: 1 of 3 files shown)


src/components/RichTextEditor.tsx
```diff
@@ -145,26 +157,27 @@ export default function RichTextEditor({
...
+          onFocus={() => {
+            setFocus(true);
+            if (onFocus) onFocus();
+          }}
+          onBlur={() => {
+            setFocus(false);
+            if (onBlur) onBlur();
+          }}
+        />
+      </Suspense>
```
htuerker, line 180: Nothing changed here, only covered with <Suspense>

There isn't any default styling, so if a problem happens on async loading it crashes. Seems impossible but I wanted to be deterministic here.

notsidney:
I noticed that when switching from light to dark, it correctly shows the dark menu. But when switching from dark to light, it’s stuck on the dark menu (the background and text colors are both white).

This is because importing the CSS file results in a `<style>` tag with the CSS to be appended to the `<head>`, but doesn’t remove any previously imported styles that have been unmounted. So the result is:
```html
<style> light styles </style>
<style> dark styles </style>
```
Then the dark styles will always take precedence from now on.

---

This can be solved by writing the `<style>` tag ourselves:

```jsx
// Editor styles
/* eslint import/no-webpack-loader-syntax: off */
import skinCss from "!!raw-loader!tinymce/skins/ui/oxide/skin.min.css";
import skinDarkCss from "!!raw-loader!tinymce/skins/ui/oxide-dark/skin.min.css";

...

<style>{theme.palette.mode === "dark" ? skinDarkCss : skinCss}</style>

```

The second problem is that the iframe used to display the content doesn’t re-render when the theme changes. The `<Editor>` component receives an updated `content_style` prop, but doesn’t update the iframe itself. We can force this to happen by setting `key={theme.palette.mode}` on the `<Editor>` component itself (which remounts the entire component).

This is ok to do because the user’s input is stored in state in the component containing `<RichTextEditor>`, so when the editor is remounted, we don’t lose the user input.

I’ll write review comments on specific lines, which you can approve yourself, then I’ll merge this PR 👍

src/components/RichTextEditor.tsx
```diff
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { lazy, Suspense, useState } from "react";
```
notsidney, line 1:
```suggestion
import { useState } from "react";
```

src/components/RichTextEditor.tsx
```diff
@@ -9,8 +9,6 @@ import "tinymce/tinymce.min.js";
 import "tinymce/themes/silver";
 // Toolbar icons
 import "tinymce/icons/default";
-// Editor styles
-import "tinymce/skins/ui/oxide/skin.min.css";
 // Content styles, including inline UI like fake cursors
```
notsidney, line 12:
```suggestion
// Editor styles
/* eslint import/no-webpack-loader-syntax: off */
import skinCss from "!!raw-loader!tinymce/skins/ui/oxide/skin.min.css";
import skinDarkCss from "!!raw-loader!tinymce/skins/ui/oxide-dark/skin.min.css";
// Content styles, including inline UI like fake cursors
```

src/components/RichTextEditor.tsx
```diff
@@ -27,6 +25,14 @@ import "tinymce/plugins/paste";
...
 import "tinymce/plugins/code";
 
+// Editor styles
+const RichTextEditorDarkCSS = lazy(
+  () => import("@src/theme/RichTextEditorDarkCSS")
+);
+const RichTextEditorLightCSS = lazy(
+  () => import("@src/theme/RichTextEditorLightCSS")
+);
+
```
notsidney, line 35:
```suggestion
```

src/components/RichTextEditor.tsx
```diff
@@ -118,15 +124,21 @@ export default function RichTextEditor({
...
-          content_style: [
-            theme.palette.mode === "dark" ? contentCssDark : contentCss,
-            theme.palette.mode === "dark" ? contentUiCssDark : contentUiCss,
-            `
+      <Suspense fallback={<></>}>
+        {theme.palette.mode === "dark" ? (
+          <RichTextEditorDarkCSS />
+        ) : (
+          <RichTextEditorLightCSS />
+        )}
```
notsidney, line 132:
```suggestion
    <style>{theme.palette.mode === "dark" ? skinDarkCss : skinCss}</style>
```

Suspense is no longer needed here.

src/components/RichTextEditor.tsx
```diff
@@ -118,15 +124,21 @@ export default function RichTextEditor({
...
+        <Editor
```
notsidney, line 133:
```suggestion
        <GlobalStyles
          styles={{
            ".tox": {
              "& .tox-menu.tox-menu, &.tox-tinymce-aux .tox-toolbar__overflow.tox-toolbar__overflow":
                {
                  backgroundColor: theme.palette.background.paper,
                  backgroundImage:
                    "linear-gradient(rgba(255, 255, 255, 0.12), rgba(255, 255, 255, 0.12))", // elevation 8
                  boxShadow: theme.shadows[8],
                  border: "none",
                  borderRadius: (theme.shape.borderRadius as number) * 2,
                },

              "& .tox-collection--list": {
                "& .tox-collection__group.tox-collection__group": {
                  padding: theme.spacing(0.5),
                },
                "& .tox-collection__item": {
                  borderRadius: theme.shape.borderRadius,
                },
              },

              "&.tox-tinymce-aux .tox-toolbar__overflow.tox-toolbar__overflow":
                {
                  padding: theme.spacing(0.5, 0),
                },
              "& .tox-tbtn.tox-tbtn": {
                borderRadius: theme.shape.borderRadius,
                margin: 0,
              },
            },
          }}
        />
        
        <Editor
```

Adding GlobalStyles here to style the dropdowns. These need to be global because they’re mounted to the DOM in a React Portal, so the `<Styles>` above (which only apply to children) won’t work here.

src/components/RichTextEditor.tsx
```diff
@@ -118,15 +124,21 @@ export default function RichTextEditor({
...
+          disabled={disabled}
```
notsidney, line 134:
```suggestion
          key={theme.palette.mode}
          disabled={disabled}
```

Added key to re-render when theme changes.

src/components/RichTextEditor.tsx
notsidney, line 180:
```suggestion
```

Suspense no longer needed

src/theme/RichTextEditorDarkCSS.tsx
```diff
@@ -0,0 +1,5 @@
+import "tinymce/skins/ui/oxide-dark/skin.min.css";
+
+export default function RichTextEditorDarkCSS() {
+  return null;
+}
```
notsidney, line 5:
```suggestion
```

You can delete this file

src/theme/RichTextEditorLightCSS.tsx
```diff
@@ -0,0 +1,5 @@
+import "tinymce/skins/ui/oxide/skin.min.css";
+
+export default function RichTextEditorLightCSS() {
+  return null;
+}
```
notsidney, line 5:
```suggestion
```

You can delete this file

notsidney (changes requested):
Please commit the suggestions. Feel free to expand the `<GlobalStyles>` to further improve styling to match our theme.

htuerker:
> This can be solved by writing the <style> tag ourselves:

I disagree. This can only be solved: scan all the entire appended stylesheets then disable invalid ones by reverse engineering. Just joking 😄 Thanks for your guidance, by the way!

htuerker:
I think this is good to go! @notsidney

https://user-images.githubusercontent.com/46192266/176274786-cb2e4983-6907-41ad-8691-8e1c1305cf60.mov

notsidney (approved):
Looks good to me!

~Also just wondering why you’re importing GlobalStyles from tss-react instead of MUI—is there an advantage there that I’m not aware of?~

Oh, I just realised I forgot to put that suggestion in. Either should be fine since we’re already using tss-react elsewhere, because of the migration from MUI v4 to v5 away from JSS.

Issue #816, opened 2022-09-16
Logic issue in action field customized icons


mshamsrainey:
Toggling off customized icons in Actions field does not revert icons to default. Icons only revert to default when customized icon is removed from input field.

**To Reproduce**
1. Create Actions field in table
2. Open the configuration settings for the column with an actions field
3. Click the checkbox to "Customize button icons with an emoji"
4. Fill in an emoji of your choice to replace the default Run icon
5. Save your changes and close the configuration
6. Re-open the configuration settings for this column
7. Un-check the checkbox to "Customize button icons with an emoji"
8. Save changes and close configuration

**Expected behavior**
Un-checking the checkbox for customized icons should revert icons back to default. Instead, the only way to revert icons to default is to remove them from the text field that appears after the customized icon checkbox is clicked.

**Screenshots**
I felt that a video explained this better than an image, so I created a screen recording that you can access at this [link](https://www.loom.com/share/bb4c9ff967a446d49aad40e8c6cb1adc)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: develop branch via localhost

**Notes:**
I'm pretty sure this is a quick logic fix, just adding in a check for whether custom icons are enabled when determining the icon to show. I ran into this while working on #513, actually, and it was a one-line fix there. I can't imagine this being substantially more complex.

mshamsrainey:
I can go ahead and work on this actually! Just wanted to make it an issue in case anyone was looking for something to play around with :-) If someone could assign me that'd be great! (cc @harinij @shamsmosowi)

Issue #513, opened 2021-09-29
Rating field - Add customization options


harinij:
**Is your feature request related to a problem? Please describe.**
Currently, the "Rating" field does not have customization.
<img width="668" alt="Screen Shot 2021-09-29 at 8 07 29 pm" src="https://user-images.githubusercontent.com/307298/135248713-a2a90437-8447-454f-9923-59fe3ad49cb5.png">

![image](https://user-images.githubusercontent.com/47289133/190283845-fb53c9b5-8907-407e-9e31-b58aef2ab02a.png)

**Describe the solution you'd like**
Allow icon/emoji picker customization in the rating field setting
- Currently the default option is star icon, allow for emoji picker option (See sample emoji picker used in the customization step of the "Action" field type)
![Screen Shot 2022-02-17 at 10 22 53 am](https://user-images.githubusercontent.com/307298/154374477-6ebef084-7a35-410f-9891-b62f9eeb198a.png)

Rating field code location: https://github.com/rowyio/rowy/tree/develop/src/components/fields/Rating

harinij:
Done @alceil 👍

alceil:
Thanks @harinij

mshamsrainey:
I'll be working on this issue as part of GHC's Open Source Day :-) @harinij

harinij:
Has been merged into dev with a recent release

PR #817, opened 2022-09-16
#513: Added Customizable Icons for Ratings Field

mshamsrainey:
#513: Added Customizable Icons for Rating Field

Demonstration here (no audio): https://www.loom.com/share/b4738b9a8a4e45e0bd9c945d9a203388 (Video demo updated to reflect side drawer and icon size updates)

Note: While working on this ticket, I encountered a bug that I created an issue for here #816. I also received some great design feedback that I wish I had time to work on, but don't think I will at the moment (at least during this one day open source challenge!).

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,9 @@ Connect to your database, manage data in table-UI with role based access control
 
 Set up Rowy on your Google Cloud Platform project with this easy deploy button.
 
-[![Run on Google Cloud](https://deploy.cloud.run/button.svg)](https://deploy.rowy.app/)
+[<img width="250" alt="Guided quick start button" src="https://user-images.githubusercontent.com/307298/185548050-e9208fb6-fe53-4c84-bbfa-53c08e03c15f.png">](https://rowy.app/)
 
-https://deploy.rowy.app/
+https://rowy.app
 
 ## Documentation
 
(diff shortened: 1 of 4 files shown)


shamsmosowi:
Hi @mshamsrainey, this looks great! 
in terms of the use of the preview fab icon button, it makes sense in the context of the action button, but here it would be more informative for the user to preview how the cell interaction is going to look

mshamsrainey:
@shamsmosowi great call, thanks for that feedback! I'll take a look at making that change either tonight or tomorrow and ping you for another review when I'm finished

mshamsrainey:
Hi @shamsmosowi, I just pushed a new commit changing the settings to give a preview of how the ratings will look using the custom icons. 

While working on this, I also realized that the existing code would allow people to leave the text field indicating what they wanted their maximum rating to be blank, or set the maximum rating to something super high like 10,000. So, I added in some simple validation to restrict users to a range 1-20 based on the input props already present in the code.

Let me know what you think!

notsidney (changes requested):
This looks good!

You should use the current `StarBorderIcon` as the default `emptyIcon`, as it’s higher contrast and improves accessibility.

image

notsidney:
@mshamsrainey Thanks! I’m merging this now and it will be part of the next release 🚀

Issue #799, opened 2022-09-13
Email field: Validation regex option


harinij:
**Is your feature request related to a problem? Please describe.**
Validation and regex options are currently available for only fields like short text and long text. Whereas the email field does not have a default validation option/customization.

If you go to the email field config [here](https://demo.rowy.io/table/allFieldTypes#page=1&columnModal=%7B%22type%22%3A%22config%22%2C%22columnKey%22%3A%22email%22%7D), you will note that validation check option is not available.

**Describe the solution you'd like**
<img src="https://user-images.githubusercontent.com/307298/189788647-6bf395c4-aa73-4317-bb69-8b506fb1d539.png" width="700"/>
To the below email config modal shown above, add the validation regex option as shown below. The validation regex option should have a default value of `[a-z0-9]+@[a-z]+\.[a-z]{2,3}`

<img src="https://user-images.githubusercontent.com/307298/189788656-c8cf2c7a-0fca-47e0-bed1-4b6e550301e5.png" width="700"/>

likitarai1:
@harinij I would like to work on this issue.

RajGM:
Hi @harinij 
I think default value could be: ^[a-zA-Z0-9+_.-]+@[a-zA-Z0-9.-]+\.[a-zA-z]{2,3}$

RajGM:
Made a PR: https://github.com/rowyio/rowy/pull/820

thinktrans:
If I may, I would like to request you to not put in the regex as decided. It invariably leaves the Internationalized Email addresses at bay. If possible, keep it as simple as this:
<e-mail local part>@<domain name>
Which can be implemented as 

For <e-mail local part> - Follow the guidelines by the "Universal Acceptance Steering Group" - UASG-026 - https://uasg.tech/download/uasg-028-considerations-for-naming-internationalized-email-mailboxes-en/

For <domain name>, you can follow any domain validation methodology using standard libraries, depending on your programming language. For the recent studies on the subject, follow the document UASG-018A - https://uasg.tech/download/uasg-018a-ua-compliance-of-some-programming-language-libraries-and-frameworks-en/.

notsidney:
Hi @thinktrans, this will be an optional feature that has to be explicitly enabled in column config.

ashutosh887:
I would like to work on this Issue!

RajGM:
Hi @ashutosh887 
I made a PR for the issues and will open a discussion about the errors I am facing.

[...]

PR #820, opened 2022-09-17
Feat:Email validation regex field added along with default values

RajGM:
Signed-off-by: Raj Gaurav Maurya <rajgmsocial19@gmail.com>

### DIFF
--- a/src/components/fields/Email/Settings.tsx
+++ b/src/components/fields/Email/Settings.tsx
@@ -1,0 +1,21 @@
+import { ISettingsProps } from "@src/components/fields/types";
+import { TextField } from "@mui/material";
+
+export default function Settings({ onChange, config }: ISettingsProps) {
+    return (
+        <>
+            <TextField
+                type="text"
+                label="Validation regex"
+                id="validation-regex"
+                value={config.validationRegex}
+                defaultValue="[a-z0-9]+@[a-z]+\.[a-z]{2,3}"
+                fullWidth
+                onChange={(e) => {
+                    if (e.target.value === "") onChange("validationRegex")(null);
+                    else onChange("validationRegex")(e.target.value);
+                }}
+            />
+        </>
+    );
+}
--- a/src/components/fields/Email/index.tsx
+++ b/src/components/fields/Email/index.tsx
@@ -13,6 +13,10 @@ const SideDrawerField = lazy(
     import("./SideDrawerField" /* webpackChunkName: "SideDrawerField-Email" */)
 );
 
+const Settings = lazy(
+  () => import("./Settings" /* webpackChunkName: "Settings-ShortText" */)
+);
+
 export const config: IFieldConfig = {
   type: FieldType.email,
   name: "Email",
@@ -26,6 +30,7 @@ export const config: IFieldConfig = {
   TableCell: withBasicCell(BasicCell),
   TableEditor: TextEditor,
   SideDrawerField,
+  settings: Settings,
   filter: {
     operators: filterOperators,
   },


RajGM:
@htuerker @harinij

notsidney:
This PR only adds the text field to save the custom regex to the column config. It does not do any validation on data. @harinij is that within the scope of the original issue (#799)?

RajGM:
Hi @notsidney 
It's working now.
Working

notsidney:
Thanks, it works when I manually update the regex (such as adding a space at the end) then saving.

The issue is that having that as the default value in the UI implies that that setting has already been set and that clicking “Update” will set it. However, reading the code, it’s only set when there is an input to the text field, so it has to be typed in first.

I don’t think it’s a good idea to have it as the default value, but rather have a button next to it that says “Use standard regex” or something similar that sets the regex. This makes sure the user doesn’t think the regex is set but it’s not working.

src/components/fields/Email/Settings.tsx
```diff
@@ -0,0 +1,21 @@
...
+                type="text"
+                label="Validation regex"
+                id="validation-regex"
+                value={config.validationRegex}
+                defaultValue="^[a-zA-Z0-9+_.-]+@[a-zA-Z0-9.-]+.[a-zA-z]{2,3}$"
+                fullWidth
+                onChange={(e) => {
+                    if (e.target.value === "") onChange("validationRegex")(null);
+                    else onChange("validationRegex")(e.target.value);
+                }}
```
notsidney, line 17: Because it’s set as `defaultValue` here, it appears as if the regex is already set. But it’s only set when the text field has its value changed, causing confusion.

notsidney (changes requested):

src/components/fields/Email/Settings.tsx
RajGM: Got it.
That means we need to have a button that fills the regex expression with the default value triggering the onChange method.

RajGM: Hi @notsidney 
Are you expecting something like this?
emailRegex

RajGM:
Hi @notsidney 

`import { ISettingsProps } from "@src/components/fields/types";
import { TextField, Button } from "@mui/material";
import { MutableRefObject, useRef } from "react";

export default function Settings({ onChange, config }: ISettingsProps) {

    const emailField = useRef<HTMLInputElement>(null);
    
    const useStandardRegex=()=>{
        console.log("Use Standard Regex triggered");
        if(emailField.current!=null){
            console.log("Not Null");
            console.log(emailField.current.value);
            emailField.current.value="^[a-zA-Z0-9+_.-]+@[a-zA-Z0-9.-]+.[a-zA-z]{2,3}$";
            console.log(emailField.current.value);
        }
    }

    return (
        <>
            <TextField
                type="text" ref={emailField}
                label="Validation regex"
                id="validation-regex"
                value={config.validationRegex}
                fullWidth
                onChange={(e) => {
                    if (e.target.value === "") onChange("validationRegex")(null);
                    else onChange("validationRegex")(e.target.value);
                }}
            />
            <Button style={{width:"200px", margin:"20px auto auto"}} onClick={useStandardRegex}>
                Use standard Regex
            </Button>
        </>
    );
}
`

emailValidation

Even though the value has been applied after clicking the button, it does not appear on the TextField.

src/components/fields/Email/Settings.tsx
RajGM: Hi @notsidney 

In the latest push https://github.com/rowyio/rowy/compare/main...RajGM:rowy:emailValidation_799

The standard regex appears as expected after clicking on the button, but it's not validating the email with regex.

src/components/fields/Email/Settings.tsx
```diff
@@ -0,0 +1,33 @@
+import { ISettingsProps } from "@src/components/fields/types";
+import { TextField, Button } from "@mui/material";
+import { useRef } from "react";
+
+export default function Settings({ onChange, config }: ISettingsProps) {
+
+    const emailFieldValidation = useRef<HTMLInputElement>(null);
+    const useStandardRegex = () => {
+        if (emailFieldValidation.current != null) {
+            emailFieldValidation.current.value = "^[a-zA-Z0-9+_.-]+@[a-zA-Z0-9.-]+.[a-zA-z]{2,3}$";
```
notsidney, line 10:
```suggestion
            onChange("validationRegex")("^[a-zA-Z0-9+_.-]+@[a-zA-Z0-9.-]+.[a-zA-z]{2,3}$");
```

As I wrote in a [previous comment](https://github.com/rowyio/rowy/pull/820#issuecomment-1264966127):

> it works when I manually update the regex (such as adding a space at the end) then saving.

> it’s only set when there is an input to the text field, so it has to be typed in first.

This is because you call `onChange` when the user types in the text field only. You manually change the value of the input element in the DOM, which doesn’t update the React state.

So when the user clicks the button, it updates the rendered input but doesn’t save in the column config, so it will never validate unless if they type in the input _after_ clicking the button.

src/components/fields/Email/Settings.tsx
```diff
@@ -0,0 +1,33 @@
+import { ISettingsProps } from "@src/components/fields/types";
+import { TextField, Button } from "@mui/material";
+import { useRef } from "react";
+
+export default function Settings({ onChange, config }: ISettingsProps) {
+
+    const emailFieldValidation = useRef<HTMLInputElement>(null);
```
notsidney, line 7:
```suggestion
```

Remove unnecessary refs.

src/components/fields/Email/Settings.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const useStandardRegex = () => {
+        if (emailFieldValidation.current != null) {
+            emailFieldValidation.current.value = "^[a-zA-Z0-9+_.-]+@[a-zA-Z0-9.-]+.[a-zA-z]{2,3}$";  
+        }
+    }
+
+    return (
+        <>
+            <TextField
+                inputRef={emailFieldValidation}
```
notsidney, line 17:
```suggestion
```

Remove unnecessary refs.

notsidney (changes requested):
Please commit the suggestions, then this PR is ready to be merged.

notsidney:
@RajGM What’s the benefit of copying to the clipboard instead of updating the value in state directly? The user would have to click the button _then_ click on the text field _then_ paste. I already provided a working suggestion that would have resolved this issue.

RajGM:
Ok, let me directly implement the changes into state using https://www.npmjs.com/package/jotai

src/components/fields/Email/Settings.tsx
```diff
@@ -0,0 +1,28 @@
+import { ISettingsProps } from "@src/components/fields/types";
+import { TextField, Button } from "@mui/material";
+
+export default function Settings({ onChange, config }: ISettingsProps) {
+
+    const copyStandardRegex = () => {
+      navigator.clipboard.writeText('^[a-zA-Z0-9+_.-]+@[a-zA-Z0-9.-]+.[a-zA-z]{2,3}$');
```
notsidney, line 7:
```suggestion
      onChange("validationRegex")("^[a-zA-Z0-9+_.-]+@[a-zA-Z0-9.-]+.[a-zA-z]{2,3}$");
```

> As I wrote in a [previous comment](https://github.com/rowyio/rowy/pull/820#issuecomment-1264966127):
> 
> > it works when I manually update the regex (such as adding a space at the end) then saving.
> 
> > it’s only set when there is an input to the text field, so it has to be typed in first.
> 
> This is because you call `onChange` when the user types in the text field only. You manually change the value of the input element in the DOM, which doesn’t update the React state.
> 
> So when the user clicks the button, it updates the rendered input but doesn’t save in the column config, so it will never validate unless if they type in the input _after_ clicking the button.

notsidney (changes requested):
No, it is entirely unnecessary to have to use Jotai. The suggestion I made previously is enough to implement this behavior. Let’s get this PR over the line.

src/components/fields/Email/Settings.tsx
RajGM: Hi Sidney
Can you please refer to some docs or something.
While I understood what you are trying to convey but being unable to implement the feature.

src/components/fields/Email/Settings.tsx
```diff
@@ -0,0 +1,28 @@
...
+                id="validation-regex"
+                value={config.validationRegex}
+                fullWidth
+                onChange={(e) => {
+                    if (e.target.value === "") onChange("validationRegex")(null);
+                    else onChange("validationRegex")(e.target.value);
+                }}
+            />
+            <Button style={{ width: "200px", margin: "20px auto auto" }} onClick={copyStandardRegex}>
+                Copy standard Regex to clipboard
```
notsidney, line 24:
```suggestion
                Use standard regex
```

notsidney (approved):

src/components/fields/Email/Settings.tsx
notsidney: I’ve gone ahead and merged this PR with some changes to make it work. An updated contribution guide will be published shortly, which will help for more complex contributions. [Here’s a draft link.](https://docs-git-v3-rowy.vercel.app/contributing/frontend-architecture)

For this feature:
1. The `Settings` component takes `onChange` and `config` as props. `config` is the current column config in the table’s schema. `onChange` will update the config in local state.
2. A parent component renders this `Settings` component and an “Update” button that will save the updated config stored in local state to the database.
3. Separately, the `TextField` component has its own `onChange` prop, where you can pass a function that will be called whenever the user types in that text field.
4. You set this to update the local state using `onChange` from `Settings`. So the user can type their own regex and it will be saved when they click “Update”.
5. When you created the “Use standard Regex” button, you did not use `onChange` from `Settings`, but rather changed the rendered input directly (using a ref). This means you did not “inform” React, so the local state did not change.
    - You can learn more about this in the React docs. We use [controlled components](https://reactjs.org/docs/forms.html#controlled-components) but you treated it like an [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html). The source of truth is in the local state (in the parent component).
6. When the user clicks “Update”, the parent component saves the local state. But since it was never updated, the standard regex is never saved to the database, so there is no validation.

The way to fix it is to use the `onChange` prop from `Settings` for the button, which you already did for the `TextField`.

Issue #523, opened 2021-10-01
Field Filter UI update - Color Field


harinij:
**Is your feature request related to a problem? Please describe.**
Currently, the Color field is not supported in the "Filter" feature on the top left of the table

![Screen Shot 2021-10-01 at 4 45 09 pm](https://user-images.githubusercontent.com/307298/135577149-0ad7cf96-e0eb-41ac-819d-38b3a3f55325.png)

**Describe the solution you'd like**
Add a color picker in the UI to allow for this field to be filterable. You can read about all the fields [here](https://docs.rowy.io/field-types/supported-fields) and how fields are structured [here](https://docs.rowy.io/field-types/add)

mshamsrainey:
Hi @harinij, I'll take a look at this!

mshamsrainey:
Hi! I have a functioning solution for this but I would love some feedback on the styling. In cells with the color field, the color picker opens up in a popover. Since the filter functionality already opens in a popover, would the ideal UI for filtering color fields be a popover inside a popover? See the below image for an idea of what the UI looks like without a separate popover for the color picker (with some clipping that I'm still working on fixing!) (cc: @harinij @shamsmosowi)
![image](https://user-images.githubusercontent.com/66425195/192106012-ba10011d-41ba-4dea-9562-65435c5a3132.png)

harinij:
Has been released

PR #894, opened 2022-10-21
#523: Added functional color field filtering

mshamsrainey:
Not sure if the UI on this is quite what y'all are looking for, but wanted to go ahead and get the PR up for feedback!


notsidney (changes requested):
This is a good start but unfortunately there’s an issue with how we store colors. The HSV values in Firestore aren’t integers, and when selecting a color using HEX or RGB, the decimal values can be different, so the filters don’t work, since it compares the whole object.

image

You can write a custom filter here [@src/hooks/useFirestoreCollectionWithAtom.ts](https://github.com/rowyio/rowy/blob/develop/src/hooks/useFirestoreCollectionWithAtom.ts#L338-L344) to only compare the `.hex` values

mshamsrainey:
Appreciate the feedback @notsidney! Just added another commit that should address your suggestion and also make a tweak to how the color field filter values are formatted, let me know what you think :-)

notsidney (approved):
Looks good to me 👍

PR #895, opened 2022-10-21
Editable Action Label

mshamsrainey:
In this PR:
- [x] Functionality to customize the label of action in part 4 of the field configuration (customization section)
- [X] Refactor to default to column name instead of key for label (per suggestion from @shamsmosowi)
- [X] UI reflecting customized action label in both side drawer and table cells


src/components/fields/Rating/Settings.tsx
```diff
@@ -1,48 +1,95 @@
...
+          label="Highest possible rating"
+          type={"number"}
+          value={config.max}
+          fullWidth
+          error={false}
+          onChange={(e) => {
+            let input = parseInt(e.target.value) || 0
+            if (input > 20) { input = 20 }
+            onChange("max")(input);
+          }}
```
notsidney, line 24:
```suggestion
          onChange={(e) => {
            let input = parseInt(e.target.value) || 0
            if (input > 20) { input = 20 }
            onChange("max")(input);
          }}
          inputProps={{ min: 1, max: 20 }}
```
I’m not sure why the Rating field changes are here, but this line needs to be put back in. I can set the max to -1 or less and it causes an error.

notsidney (changes requested):

src/components/fields/Rating/Settings.tsx
notsidney: @mshamsrainey The rest looks good to me so I’ll commit this myself and merge now. Just wanted to make sure you see this

notsidney (approved):

PR #903, opened 2022-10-28
Refactor on custom rating icons

mshamsrainey:
To mitigate issues with "empty" and standard custom icons looking the same in browsers running on WebKit, I created an Icon component using the logic written in previous PRs that I could then modify to show "empty" custom icons with an explicitly-declared opacity value. No changes to the UI on Chrome, Edge, or Firefox--the only intended change would be for WebKit browsers.


src/components/fields/Rating/Icon.tsx
```diff
@@ -0,0 +1,25 @@
+import RatingIcon from "@mui/icons-material/Star";
+import RatingOutlineIcon from "@mui/icons-material/StarBorder"
+import { get } from "lodash-es";
+
+
+export default function Icon(config: any, isEmpty: Boolean) {
```
notsidney, line 6:
```suggestion
export interface IIconProps {
  config: any;
  isEmpty: boolean;
}

export default function Icon({ config, isEmpty }: IIconProps) {
```
In React components, the first argument is an object of props. You have `config` and `isEmpty` as separate arguments. In this case, `isEmpty` isn’t actually a prop you can pass to the component using `<Icon isEmpty={ ... }>`. So on non-emoji Rating fields, this component always runs `getStateIcon`:
image

notsidney (changes requested):
The issue is that MUI uses different text colors with opacity, `hsla(..., opacity)` to make it appear not filled in. But WebKit browsers like Safari don’t honor the opacity in text color, and always display emojis with full opacity. (React Native had a similar issue a while back: https://github.com/facebook/react-native/issues/2241)

You can change the global theme component style instead to use opacity and use the same text color as filled icons: https://github.com/rowyio/rowy/blob/enhancement/513-ratings-customized-icons/src/theme/components.tsx#L1376

```ts
      MuiRating: {
        styleOverrides: {
          icon: {
            // https://github.com/mui/material-ui/issues/32557
            "& .MuiSvgIcon-root": { pointerEvents: "auto" },
            color: theme.palette.text.secondary,
          },
          iconEmpty: { opacity: 0.38 },
        },
      },
```

Then you also don’t need to make `Icon` into a separate component and keep the code simple like previously.

mshamsrainey:
@notsidney good catch! I hadn't thought to check the issues on React Native. Let me know how this new version looks to you (it's also got some package.json changes that I believe are just a result of merging main into my branch, shouldn't be anything too exciting)

src/components/fields/Rating/Icon.tsx
```diff
@@ -0,0 +1,28 @@
...
+import RatingOutlineIcon from "@mui/icons-material/StarBorder"
+import { get } from "lodash-es";
+
+
+export interface IIconProps{
+  config: any,
+  isEmpty: boolean
+}
+
+export default function ({config, isEmpty} : IIconProps) {
```
notsidney, line 11:
```suggestion
export default function Icon({config, isEmpty} : IIconProps) {
```
Please name function components, so they appear correctly in React DevTools.

notsidney (approved):
Thanks @mshamsrainey.

Please don’t merge from main, as it’s a stale branch. Since the PR is into develop, please merge from develop instead. It appears that the package.json and yarn.lock updates downgrade and remove some dependencies, which have prevented the app from running on my machine.

I also want to make the point that you’re not expected to look at React Native issues since we don’t use React Native at all. I linked to that issue as an example, after Googling the primary issue of text color.

PR #913, opened 2022-11-02
Changed default Action to script from cloud func

mshamsrainey:
This PR modifies the settings for the Action field to default to a user-declared script instead of a cloud function that they've already deployed to Firestore/GCP.

### DIFF
--- a/src/components/fields/Action/Settings.tsx
+++ b/src/components/fields/Action/Settings.tsx
@@ -351,7 +351,7 @@ const Settings = ({ config, onChange, fieldName }: ISettingsProps) => {
                 </RadioGroup>
               </FormControl>
 
-              {config.isCloudFunction ? (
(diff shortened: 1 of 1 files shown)


notsidney (changes requested):
Hi @mshamsrainey, you can’t change fields stored in the config without breaking backwards compatibility for some users. In this case, if they had clicked the “Cloud Function” radio button, it would have set `config.isActionScript` to `false`. Then, after this change, `config.isCloudFunction` would be `undefined`, so it displays the script options.

Currently, if `config.isActionScript` is falsy (`false` or `undefined`), it defaults to showing the Cloud Function options. You can alternatively check if it was explicitly set to `false`, so the default value, `undefined`, displays the script options.

notsidney (approved):

PR #918, opened 2022-11-04
Enhancement/show total rows in db

mshamsrainey:
This PR implements the new Firebase function getCountFromServer and stores its data in a Jotai atom with table scope to have the row count in the table status bar reflect the number of documents in the Firestore collection.


shamsmosowi:
> **The latest updates on your projects**. Learn more about [Vercel for Git ↗︎](https://vercel.link/github-learn-more)
> 
> Name	Status	Preview	Updated
> **rowy-os**	❌ Failed ([Inspect](https://vercel.com/rowy/rowy-os/FSMVv5L1qPzAkFa156UuDPaoDwYS))		Nov 4, 2022 at 7:50PM (UTC)

@mshamsrainey are you able to run `yarn build` command successfully locally?

mshamsrainey:
@shamsmosowi yes, I did run into that build error but fixed it in my most recent commit--I don't have access to rerun the Vercel deployment, but after that commit it builds locally at least.

src/hooks/useFirestoreCollectionWithAtom.ts
```diff
@@ -191,6 +196,10 @@ export function useFirestoreCollectionWithAtom<T = TableRow>(
               available: docs.length >= memoizedQuery.limit,
             }));
           }
+          // on each new snapshot, use the query to get and set the document count from the server
+          getCountFromServer(memoizedQuery.query).then((value) => {
```
shamsmosowi, line 200: the `memoizedQuery.query` already has a limit apply to it causing the getCountFromServer function only returns the number of total loaded documents, rather than total number documents in the collection. I'm not sure if you can remove the limit from `memoizedQuery.query` otherwise you can just recreate the query with collection path and filters only

shamsmosowi:
hey @mshamsrainey, it mostly work other than the change requested in the comment, 
here's an example of the intended outcome 
Screen Shot 2022-11-08 at 10 30 27 pm

notsidney (approved):

src/hooks/useFirestoreCollectionWithAtom.ts
```diff
@@ -190,6 +195,10 @@ export function useFirestoreCollectionWithAtom<T = TableRow>(
               available: docs.length >= memoizedQuery.limit,
             }));
           }
+          // on each new snapshot, use the query to get and set the document count from the server
+          getCountFromServer(memoizedQuery.unlimitedQuery).then((value) => {
+            setServerDocCountAtom(value.data().count)
+          })
```
notsidney, line 201:
```suggestion
          // on each new snapshot, use the query to get and set the document count from the server
          if (serverDocCountAtom) {
            getCountFromServer(memoizedQuery.unlimitedQuery).then((value) => {
              setServerDocCountAtom(value.data().count)
            })
          }
```
⚠️ `serverDocCountAtom` is an optional prop and we don’t query the total number of docs by default. As said in the comment on lines 107–108, you need to make sure that that prop was passed in the first place.

Currently, if the component calling this hook does not pass that atom, it will overwrite the `dataAtom` with the number of docs in the query, breaking the hook entirely.

We need to pass any atom to `useSetAtom`, otherwise Jotai will throw an error. So that’s why we need to be very careful when making a modification to a hook in this way.

notsidney (changes requested):

notsidney (approved):

PR #983, opened 2022-11-24
#562: Add accessible reorder option in column menu

iamanishroy:
[#562](https://github.com/rowyio/rowy/issues/562): Add accessible reorder option in column menu.

Worked on this improvement.

I started by creating a new file ManageColumns.tsx and keeping HiddenFields.tsx as the base.

I have used [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd) as recommended.
Dragging will be disabled when the search text field is not empty.

**Note**: HiddenFields.tsx is still there for reference.

image

Issue #562, opened 2021-10-15
Add accessible reorder option in column menu


notsidney:
**Is your feature request related to a problem? Please describe.**
Columns can only be reordered using drag and drop, but this is not an accessible method and moving the column from one end of the table to another is difficult if there are many columns.

**Describe the solution you'd like**
In the column menu, there should be an option to reorder that opens a modal to change the order of columns.

gibsonliketheguitar:
I would like to claim this task.

harinij:
This issue is open, needs design.

Some suggested options, the menu for column can be renamed to "Columns" 
<img width="577" alt="Screen Shot 2022-07-05 at 9 57 23 pm" src="https://user-images.githubusercontent.com/307298/177322591-edb19824-cd77-442c-8613-cd0f2a786c0b.png">

and can allow for drag and drop reorder as shown below (example, not final design suggestion). 

https://user-images.githubusercontent.com/307298/177322742-bca137fc-5bed-4992-9edc-780542797bec.mov

++ @tinacai97

notsidney:
We use [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd) already and it should be used here too as it’s accessible for keyboard and screen reader users

iamanishroy:
I am working on it

shamsmosowi:
Hi @iamanishroy , search is need specially when tables get big, 
why is it necessary to remove search or have separate menu?
you can elaborate on the blocker you're facing

shamsmosowi:
oh, i see, i think the solution would that the ordering feature is disabled if a filter is applied

harinij:
Merged to develop

PR #984, opened 2022-11-25
#562: Add accessible reorder option in column menu [new]

iamanishroy:
[#562](https://github.com/rowyio/rowy/issues/562): Add accessible reorder option in column menu.

Worked on this improvement.

I started by creating a new file ManageColumns.tsx and keeping HiddenFields.tsx as the base.

I have used [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd) as recommended.

- [x] Ordering feature will be disabled when the search filter is applied.
- [x] No changes on the appearance of button.

image

_Linked to PR [#983 ](https://github.com/rowyio/rowy/pull/983) -> created because of some issue with branch._

### DIFF
--- a/src/components/TableToolbar/HiddenFields.tsx
+++ b/src/components/TableToolbar/HiddenFields.tsx
@@ -161,6 +161,7 @@ export default function HiddenFields2() {
     });
   }
 
+  // checks whether to disable reordering when search filter is applied
   function checkToDisableDrag(e: ChangeEvent<HTMLInputElement>) {
     setDisableDrag(e.target.value !== "");
   }
--- a/src/components/TableToolbar/HiddenFields.tsx
+++ b/src/components/TableToolbar/HiddenFields.tsx
@@ -102,7 +102,13 @@ export default function HiddenFields() {
             >
               <DragIndicatorOutlinedIcon
                 color="disabled"
-                sx={[{ marginRight: "6px", opacity: disableDrag ? 0.6 : 1 }]}
+                sx={[
+                  {
+                    marginRight: "6px",
+                    opacity: (theme) =>
+                      disableDrag ? theme.palette.action.disabledOpacity : 1,
+                  },
+                ]}
               />
             </Box>
             <ColumnItem option={option}>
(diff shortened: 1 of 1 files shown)


src/components/TableToolbar/HiddenFields.tsx
```diff
@@ -67,49 +88,113 @@ export default function HiddenFields() {
...
+      >
+        {(provided) => (
+          <li {...props} ref={provided.innerRef} {...provided.draggableProps}>
+            <Box
+              sx={[{ position: "relative", height: "1.5rem" }]}
+              {...provided.dragHandleProps}
+            >
+              <DragIndicatorOutlinedIcon
+                color="disabled"
+                sx={[{ marginRight: "6px", opacity: disableDrag ? 0.6 : 1 }]}
```
notsidney, line 105:
```suggestion
                sx={[
                  {
                    marginRight: "6px",
                    opacity: (theme) =>
                      disableDrag ? theme.palette.action.disabledOpacity : 1,
                  },
                ]}
```

There isn’t enough difference between the disabled and enabled icon’s opacities.

notsidney (approved):
Thanks @iamanishroy, this looks good to me. It looks like you had to read the docs of https://github.com/rowyio/multiselect to implement this as well, so thanks for being diligent.

### A note on extending this menu in the future

This was built with the MultiSelect component, which was designed specifically for multiple selection. Adding more functionality like this would be better served by **not using MultiSelect for this anymore**.

There are two “hacks” that have been needed to implement this feature:

1. Create a new `dragDisabled` state to track if the user has input anything in the search box. This is done by tracking the `onInput` event. This should be derived from the state of that input’s value, but this is intentionally not accessible in the MultiSelect component.
2. There are two paths to access items with the keyboard. When the input is focused, the user can press the arrow keys to select columns to be hidden. This is consistent with the [WAI-ARIA guidelines](https://mui.com/material-ui/react-autocomplete/#accessibility) implemented by the MUI Autocomplete component, on which this component is based.
  But to reorder columns, the user must press Tab to select each individual column. This also makes the “Clear” and “Done” buttons less accessible, as previously they only had to Tab once to get to them, but must now go through all columns.

Issue #699, opened 2022-07-05
Prevent VIEWER role to export CSV


aritraroy:
**Is your feature request related to a problem? Please describe.**
Currently if I set VIEWER role and set this option - "Read-only for non-ADMIN users", the user is not able to add/edit/remove the database which is good. But the user is able to export the entire database in one go.

I understand that exporting the database as CSV can still be considered a view-only option, but its a security risk for our usecase and data. Any of the users whom we even grant VIEW only access, can download the entire database and take it with them as a CSV. 

**Describe the solution you'd like**
As a simple solution to this, it would have been great if there was an option to disable or hide the "CSV export" option from "Access Control" panel. This would make our data much more secure as users can only view it inside Rowy and not be able to take the entire data with them in one go.

shamsmosowi:
Hi @aritraroy, I like the idea of an access control panel, it could help with controlling access to other features as well.
My main concern with providing it for CSV export, is that it can provide a false sense of security, since users can still access all the data that they can export, either manual or if its a technical user they could use the browser console to write a script that can achieve the same result as the csv export. 
I do understand both of those options are mostly not feasible for your user base

aritraroy:
Yes, I understand it is always possible to extract the data by writing a script or through other means. But it would be impossible for non-technical users. The idea is to not make it so easy as to just download it in once click and get the entire database in one go.

We are unable to use it properly and give access to it to larger team just because of this concern.

aritraroy:
@shamsmosowi Can this feature be considered? If not, please let me know so that I can close the issue.

aritraroy:
@shamsmosowi @notsidney Haven't heard back on this for a while. Closing this issue.

notsidney:
Hi, this is a useful feature that we could look into in a future sprint for advanced access controls. I’m reopening this issue.

aritraroy:
@notsidney Sure. Any tentative timeline on when this will be available?

notsidney:
Our current focus is to make it easier for users to get started with Rowy. We cannot provide any timeline on this issue. If we can, we usually include that information in the replies to issues.

RajGM:
Hi @notsidney 
I was able to disable the export button, but I cannot access the user's auth level.

I imported { userRolesAtom,currentUserAtom } from "@src/atoms/globalScope/auth"; but it doesn't contain the user auth Level.
Is there any other function that allows us to do so?

RajGM:
Hi @notsidney 
Made PR for the requested feature.

PR #988, opened 2022-11-30
Prevent viewer export 699

RajGM:
Implemented the features as requested in https://github.com/rowyio/rowy/issues/699
Users with only VIEWER roles cannot export the data now.


harinij:
We are refactoring access control at table level to support this and this change will not fit into the new structure. So closing this out for now.

RajGM:
Got it.

> We are refactoring access control at table level to support this and this change will not fit into the new structure. So closing this out for now.

PR #1023, opened 2022-12-17
ROWY-816 Fix: Small blip in the ordering columns

iamanishroy:
### Issue
1. Try to reorder columns via drag and drop
2. There is a small blip in the ordering as showing in the video below

issue-gif

### Fix

- After close inspection, I found the cause of this issue lies with updating the state, as it's taking a noticeable time to update. Thus, resulting in this flicker effect.
- A possible fix I came up with is performing an optimistic update before truly updating the state.

And the final result.

fixed-gif


src/components/TableToolbar/HiddenFields.tsx
```diff
@@ -157,9 +163,32 @@ export default function HiddenFields() {
 
   const updateColumn = useSetAtom(updateColumnAtom, tableScope);
 
+  // function to revalidate quickAccessTableColumnsOrdered
```
shamsmosowi, line 166: https://user-images.githubusercontent.com/13803906/209665377-92ad4420-d9c9-4113-88b2-6e0a905c9cb8.mov

While this solution works for reordering items on the hidden columns UI, it doesn't solve the same issue occurring on the table DnD interaction, it also adds unnecessary computation of reordering that's already being handled by the updateColumn atom, I think an ideal solution would be to implement optimistic updates in the `useFirestoreDocWithAtom` and solve all similar issues

shamsmosowi (changes requested):

src/components/Table/TableCell/EditorCellController.tsx
```diff
@@ -53,17 +56,20 @@ export default function EditorCellController({
   }, [isDirty, localValueRef, setLocalValue, value]);
 
   // This is where we update the documents
-  const handleSubmit = () => {
+  const handleSubmit = async () => {
```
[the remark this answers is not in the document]
iamanishroy: Okay

src/components/TableToolbar/HiddenFields.tsx
iamanishroy: As suggested, I worked on it.

shamsmosowi (approved):

PR #1026, opened 2022-12-18
Better derivative evaluation out of memory error message

bofeiw:
Derivative evaluation might fail due to Rowy Run out of memory.

Instead of "failed to fetch", 
Screenshot 2022-12-19 at 04 36 24

display the issue and ask user to increase memory in GCP console.
Screenshot 2022-12-19 at 04 38 24

### DIFF
--- a/src/components/fields/Derivative/ContextMenuActions.tsx
+++ b/src/components/fields/Derivative/ContextMenuActions.tsx
@@ -2,13 +2,17 @@ import { useAtom } from "jotai";
 import { find, get } from "lodash-es";
 import { useSnackbar } from "notistack";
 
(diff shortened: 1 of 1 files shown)


shamsmosowi:
`failed to fetch` can potentially caused by multiple reasons, would it be possible for the server to return an `out of memory` error or something more specific?

bofeiw:
@shamsmosowi Unfortunately, that error message is out of our control. GCP cuts the execution off, leaving client side to have this `failed to fetch` error.

If the error is not caused by GCP cutting the execution off, we have the control and we can customise the UI. Let's keep an eye on it.

PR #1029, opened 2022-12-19
Cloud build configuration fix

mtrofimiuk:
The build was failing. I suggest using `lts` node version for the build. 

Firebase deployment steps should be run with firebase builder. This may require an update to documentation (if there's any) since the builder needs to be built for each GCP project.  (https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/firebase)


harinij:
Closing this as it is out of scope with the current upgraded codebase on develop branch.

PR #1042, opened 2022-12-30
ROWY-844 Feat: Import JSON

iamanishroy:
image

### High Level Flow
Accept JSON -> convert JSON to CSV -> parse CSV

### DIFF
--- a/src/atoms/tableScope/ui.ts
+++ b/src/atoms/tableScope/ui.ts
@@ -109,7 +109,7 @@ export type ImportAirtableData = { records: Record<string, any>[] };
 
 /** Store import CSV popover and wizard state */
 export const importCsvAtom = atom<{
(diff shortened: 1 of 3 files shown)


shamsmosowi (changes requested):
Looks good, I like the idea of converting json to csv before processing it, however it looks like its having issues detecting some types, like here its picking up a json array as url type
Screenshot 2022-12-30 at 10 31 54 am

iamanishroy:
This problem persists when we import from CSV.
I believe a new issue can be opened for the fix.

image

shamsmosowi:
> Looks like this problem persists when we import from CSV also. So, I believe a new issue can be opened for the fix.
> 
> image

You're right, looks like its not causing it, we can look into this separately, I think @mshamsrainey might be working on something

Issue #959, opened 2022-11-19
New column type for Array - To be displayed as with subtable view


shamsmosowi:
Add  a new column type: **ArrayTable**

This is a repeated ask from users of Rowy to support array fields of Firestore, for example [here](https://github.com/rowyio/rowy/discussions/956)

**Task:**
- Create a new column type called Array: Documentation for adding a new column is available [here](https://docs.rowy.io/contributing/add-a-field-type).
- This Array field on the Rowy Table UI should open similarly to how [subtable](https://demo.rowy.io/table/allFieldTypes/subTable/allFieldTypes%2F08np1ItulUrTjNTmfLNU/subtable?parentLabel=This%20is%20a%20test%20for%20action%20%20dhad%20dadlfkj) column type shows up.
- However on the Firestore side, the data should be stored as an array as shown below and not a subcollection (subtable field is stored as a subcollection).
<img width="615" alt="Screen Shot 2022-12-30 at 1 02 27 am" src="https://user-images.githubusercontent.com/307298/209964191-dfe24a3f-b2fe-441c-b930-302638f362f8.png">
<img width="547" alt="image" src="https://user-images.githubusercontent.com/53828745/202520232-675334a2-8d99-4852-8a87-d9d007abde36.png">
</div>

S-P-Ks:
I can take this task.

harinij:
Hi @S-P-Ks I have assigned the task to you. Please join us on [discord](https://www.rowy.io/discord) if you like any additional information or want to discuss technical details under the #contribution channel.

ashutosh887:
HI @harinij 
Is this Task resolved?

ashutosh887:
Thanks for assigning me!
Can you tell me how to get started with Project @shamsmosowi !

harinij:
@ashutosh887 since you are already assigned #519, I am going to unassign this task from you and reassign it to someone else as per our contribution guidelines

jyash97:
Hey 👋 

I am working into this issue, and have setup the project locally and looked into how the to add the column. 

But not sure how do we want to render the UI, if you can give more context on the UI of editing and displaying the cell value. Do we want the array to be typed ? Like shown in the screenshot ?

shamsmosowi:
hi @jyash97, 
To give you more context, we already have column type called [subtable](https://docs.rowy.io/field-types/sub-table), which renders a table for specified subcollection. The idea here is to have the same experience but for an array of objects in same document rather than in a subcollection.
[...]

PR #1045, opened 2022-12-30
feat: add array column type

jyash97:
Potential fix for #959

### DIFF
--- a/src/components/fields/Array/SideDrawerField.tsx
+++ b/src/components/fields/Array/SideDrawerField.tsx
@@ -1,0 +1,233 @@
+import { ISideDrawerFieldProps } from "@src/components/fields/types";
+
+import {
+  Stack,
+  Box,
+  Button,
+  Select,
+  MenuItem,
+  FormControlLabel,
+  TextField,
+  Switch,
+  Chip,
+  Badge,
+} from "@mui/material";
+import AddIcon from "@mui/icons-material/Add";
+import ClearIcon from "@mui/icons-material/Clear";
+
+import JSONField from "@src/components/fields/Json/SideDrawerField";
+import CodeEditor from "@src/components/CodeEditor";
+
+function ArrayFieldInput({
+  onChange,
+  value,
+  disabled,
+  id,
+}: {
+  onChange: (value: any) => void;
+  value: any;
+  disabled: boolean;
+  id: string;
+}) {
+  switch (typeof value) {
+    case "bigint":
+    case "number": {
+      return (
+        <TextField
+          onChange={(e) => onChange(+e.target.value)}
+          type="number"
+          placeholder="Enter value"
+          disabled={disabled}
+          id={id}
+        />
+      );
+    }
+
+    case "string": {
+      return (
+        <TextField
+          onChange={(e) => onChange(e.target.value)}
+          type="text"
+          placeholder="Enter value"
+          disabled={disabled}
+          id={id}
+        />
+      );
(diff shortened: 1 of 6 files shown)


src/components/fields/Array/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  type: FieldType.array,
+  name: "Array",
+  group: "Code",
+  dataType: "object",
+  initialValue: undefined,
+  initializable: true,
+  icon: <DataArrayIcon />,
+  settings: Settings,
+  description:
+    "Connects to a sub-table in the current row. Also displays number of rows inside the sub-table. Max sub-table depth: 100.",
```
jyash97, line 26: Need description

src/components/fields/Array/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  TableCell: withRenderTableCell(DisplayCell, SideDrawerField, "popover", {
+    popoverProps: { PaperProps: { sx: { p: 1, minWidth: "200px" } } },
```
jyash97, line 28: I went with the popover approach because the sub-table approach opens a Modal from the bottom, and thus the user loses the context of the row since he has to close the modal to see the row and the Modal breaks the flow, IMO Popover approach seems to be simple steps to go with. Let me know I can use modal in case this isn't what we want

src/components/fields/Array/SideDrawerField.tsx
```diff
@@ -0,0 +1,233 @@
...
+          <CodeEditor
+            defaultLanguage="json"
+            value={JSON.stringify(value) || "{\n  \n}"}
+            onChange={(v) => {
+              try {
+                if (v) onChange(JSON.parse(v));
+              } catch (e) {
+                console.log(`Failed to parse JSON: ${e}`);
+              }
+            }}
```
jyash97, line 81: This can be used to add array of objects

jyash97:
Need to add delete field option

shamsmosowi:
Hi @ jyash97, this column type is an interesting idea. It could be helpful in some cases, especially if we add support for Rowy column types, not just the base types.
I don't think it solves the main objective of #959, which is mainly concerned with the need to provide users with a way to enter arrays of structured JSON data easily/safely

src/components/fields/Array/index.tsx
shamsmosowi: You can try embedding the table component inside the popover, however providing a way to configure the column schema might be more challenging

jyash97:
@shamsmosowi 

Thanks for the review. 

I think the issue needs to be more informative since it doesn't focus on the structured JSON data and just mentions array fields. Now I have a better understanding of what the issue is about, tho I have a few questions:

1. Where to store the structure of the array, let's say the structure should be `{title: string; description: string}[]`, where we would save it & provide an option to configure it? 
2. On opening the drawer from the bottom, where do we show the schema? Or we would just show the data 
3. Which types to use for the table and will it be editable? string, boolean? or the Short text, long text types?
4. Will adding a new row save the data automatically ? Or do we need to display a save button option to save the new rows?
5. How to render a nested object in the array table? Let's say we have a structure: 
    ```
    meta: { name: string; id: string; }
    ```
     So we render the column as `meta.name` or do we just render `meta` column with JSON editor ?

jyash97:
Closing this because of inactivity and the PR does not solve the issue of adding schema and adding/editing an array of object.

PR #1046, opened 2022-12-30
chore: add nvmrc file

jyash97:
this helps in auto detect scripts or if we use `nvm use` it reads specific version from this file


jyash97:
Closing this because of inactivity

PR #1047, opened 2022-12-30
fix: prevent redirecting to link when deleting file

jyash97:
When deleting file from `type=file` it takes to the link since we use `Link`, preventing default fixes the behavior


htuerker (approved):
Hey @jyash97, It looks good to me!

Thanks for the contribution.

PR #1052, opened 2023-01-03
fix: optimistic update

iamanishroy:
used `useUpdateAtom` to fix [this](https://github.com/rowyio/rowy/pull/1048).

### DIFF
--- a/src/hooks/useFirestoreDocWithAtom.ts
+++ b/src/hooks/useFirestoreDocWithAtom.ts
@@ -161,14 +161,7 @@ export function useFirestoreDocWithAtom<T = TableRow>(
       // reset the atom’s value to prevent writes
       if (updateDataAtom) setUpdateDataAtom(undefined);
     };
-  }, [
-    memoizedDocRef,
-    updateDataAtom,
-    setUpdateDataAtom,
-    enqueueSnackbar,
-    setDataAtom,
-    disableSuspense,
-  ]);
+  }, [memoizedDocRef, updateDataAtom, setUpdateDataAtom, enqueueSnackbar]);
 }
 
 export default useFirestoreDocWithAtom;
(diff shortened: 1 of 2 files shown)


src/hooks/useFirestoreDocWithAtom.ts
```diff
@@ -60,7 +61,7 @@ export function useFirestoreDocWithAtom<T = TableRow>(
   } = options || {};
 
   const [firebaseDb] = useAtom(firebaseDbAtom, projectScope);
-  const setDataAtom = useSetAtom(dataAtom, dataScope);
+  const setDataAtom = useUpdateAtom(dataAtom, dataScope);
```
htuerker, line 64: `useUpdateAtom` and `useSetAtom` are the same, for to stick with the updated version let's continue using `useSetAtom` 

https://jotai.org/docs/utils/use-update-atom

> For newer versions of Jotai you can also use [useSetAtom](https://jotai.org/docs/api/core#use-set-atom) instead.

src/hooks/useFirestoreDocWithAtom.ts
```diff
@@ -148,6 +149,13 @@ export function useFirestoreDocWithAtom<T = TableRow>(
           }
         }
 
+        // using `setDataAtom` here we are quickly changing state to instantly update the UI
+        // after that on actual update `setDataAtom` called once again with the final data
+        // the fix is mainly to fix the flicker effect on column reordering
+        setDataAtom((prev) => {
```
htuerker, line 155: https://github.com/rowyio/rowy/blob/2f8a3144223aee4ff986148596d9299c23ada566/src/hooks/useFirestoreDocWithAtom.ts#L78-L84

As shown in the first `useEffect`, we set `dataAtom` with a Promise to achieve React suspend. That was the root cause of why it got stuck with a loading spinner when we pushed this code to the staging branch. 

Solution is simple, just add a guard here to not set `dataAtom` in cases we want to suspense:
```typescript
if (disableSuspense) {
   setDataAtom((prev) => ({ ...prev, ...updateToDb }));
}
```

htuerker (changes requested):

src/hooks/useFirestoreDocWithAtom.ts
iamanishroy: @htuerker, thanks for the explanation.

htuerker:
Hey @shamsmosowi, this PR will cause merge conflicts when you do upstream merge. I'm going to create another mirror PR from this one. It'll be easier for me to test both sides. However, please inform me when you merge, or let me handle those merges.

htuerker (approved):

PR #1056, opened 2023-01-05
ROWY-831: Add preview table

htuerker:


shamsmosowi:
Works perfectly!
Are we able to hide the following table features from the preview mode
- the drop-down for different add row types can be restricted to add to the top only
- DnD column headers
- sorting column button
- column settings button
- row menu button
https://user-images.githubusercontent.com/13803906/210782991-9cb5bce5-7421-4534-8951-a819a0c87fcc.mov

htuerker:
> Works perfectly! Are we able to hide the following table features from the preview mode
> 
> * the drop-down for different add row types can be restricted to add to the top only
> * DnD column headers
> * sorting column button
> * column settings button
> * row menu button
> 
>  Screen.Recording.2023-01-05.at.1.41.00.pm.mov

https://github.com/rowyio/rowy/blob/ROWY-831-preview-table/src/components/fields/Formula/PreviewTable.tsx#L38

There're already some hidden parts. For now, it's good to use CSS to hide things in the Table view. But later on, we may want to build a generalized solution such as using toolset variable: ["filters", "row_actions", "column_settings"]

shamsmosowi:
> > Works perfectly! Are we able to hide the following table features from the preview mode
> > 
> > * the drop-down for different add row types can be restricted to add to the top only
> > * DnD column headers
> > * sorting column button
> > * column settings button
> > * row menu button
> > 
> > Screen.Recording.2023-01-05.at.1.41.00.pm.mov
> 
> https://github.com/rowyio/rowy/blob/d80815f1d837438bbb073b594ef4191ca536a4c6/src/components/fields/Formula/PreviewTable.tsx#L38-L66
> 
> There're already some hidden parts. For now, it's good to use CSS to hide things in the Table view. But later on, we may want to build a generalized solution such as using toolset variable: ["filters", "row_actions", "column_settings"]
yep, that's fine lets hide the rest with css as well

PR #1061, opened 2023-01-06
ROWY-818: requires cloud function setup before creating related columns

bofeiw:
Currently if I continue to use Derivative without Rowy Run, evaluate will always show "cell evaluated" and function builder is always stuck at pending.

Screen Shot 2022-12-09 at 2 28 22 am

This PR requires cloud function setup before user is able to create the column by disabling the option and ask for setup.

https://user-images.githubusercontent.com/34177142/211042459-f04b529d-eeff-49ef-b732-2f6381721512.mov


bofeiw:
Updates: default values and columns are consistent now
Screenshot 2023-01-07 at 05 31 21
Screenshot 2023-01-07 at 05 33 24

PR #1090, opened 2023-01-22
ROWY-915: Fixed numbers getting inverted

iamanishroy:
image


shamsmosowi:
looks like this was included in the other pr

PR #1103, opened 2023-02-01
ROWY-444: Save sort on the table

iamanishroy:
It will ask if this sorting should be applied to all users when we sort by a field.
If yes, then every time that table opens then the sorting will be applied by default.

image


shamsmosowi:
can we have an option to allow users to save the filter just to them self, similar to filters?

PR #1112, opened 2023-02-08
Reorder image file field

saravanan-inc:


shamsmosowi (changes requested):
Hi @saravanan-inc thank you for the contribution,this looks great!
two things to improve on would be:
- add support for reordering in the sidedrawer
- hide the drag and drop icon when there's only one image in the cell, to save on space

saravanan-inc:
Hi @shamsmosowi thanks for updating on this issue and the bug,

Added support for reordering the both the image and file fields in the SideDrawer
Add hide the image and file fields drag option if there is only one file or one image

shamsmosowi:
> Hi @shamsmosowi thanks for updating on this issue and the bug,
> 
> Added support for reordering the both the image and file fields in the SideDrawer Add hide the image and file fields drag option if there is only one file or one image

Perfect! Thank you @saravanan-inc !

PR #1152, opened 2023-03-10
IROWY-935: Support import files

iamanishroy:
// draft - requires testing

image


shamsmosowi:
@iamanishroy do these changes apply to airtable import or only csv?

iamanishroy:
Right now it only supports CSV and JSON.

Issue #1218, opened 2023-04-20
Bug: Formula field value does not show up on the side panel


harinij:
**Describe the bug**
Formula field value does not show up on the side panel
![image (1)](https://user-images.githubusercontent.com/307298/233309434-9b6b2685-e205-46b4-b12b-755b6fd321e3.png)

**To Reproduce**
1. Create a formula field
2. See the value in the table
3. Does not show up in the side panel

**Expected behavior**
A formula field value should be shown in both the main table and the side panel

mnmt7:
@harinij I would like to work on this.

jemiluv8:
@mnmt7 I've started working on it.

mnmt7:
Hello @jemiluv8, I already made significant progress towards resolving the issue when I asked for the assignment and was just waiting for the assignment of the issue before creating a pull request.

However, I have a slight different way of implementing the solution for this issue. Specifically, I have taken care to ensure that the appropriate JSX component is rendered for each of the fields. While the generic TextField component may be suitable for Number and Text fields, it may not be the best solution for more complex fields such as Rating and Slider.

![screenshot](https://user-images.githubusercontent.com/107614965/233971322-6fcb585d-f54c-4fda-a3d1-1fd01414fb72.png)

By tailoring the rendered component to the specific field type, I believe that we can achieve a more efficient and effective solution to the issue.

jemilu2:
@mnmt7  I agree. Your solution is superior. Kudos.

I kinda always assumed formula fields will always 
return something that is a number or text. I didn't look 
into the other "Field Types" that may be assigned them.

mnmt7:
@jemilu2 I was of the opinion that since Lodash .isEqual() performs deep comparision, it will be automatically handled. 
`if (!isEqual(prev.field, next.field)) return false;`
However, I am open to the possibility that my assumption may have been incorrect.

harinij:
Hi @mnmt7 and @jemilu2 thanks to you both for looking into this and submitting PR, appreciate the effort. Given that @mnmt7 had commented here first for the assignment, I have assigned the task now and we will be reviewing the PR and adding feedback soon.

harinij:
Thanks @mnmt7 for your contribution 🎉

PR #1225, opened 2023-04-24
Implement and Add SideDrawerField for Formula Field

mnmt7:
This change resolves the issue #1218 by introducing a new SideDrawerField component specifically for the Formula field and making appropriate changes to the codebase to utilize this new component.

screenshot

/claim #1218

### DIFF
--- a/src/atoms/tableScope/rowActions.ts
+++ b/src/atoms/tableScope/rowActions.ts
@@ -354,7 +354,9 @@ export const updateFieldAtom = atom(
 
     const row = find(tableRows, ["_rowy_ref.path", path]);
     if (!row) throw new Error("Could not find row");
-    const isLocalRow = Boolean(find(tableRowsLocal, ["_rowy_ref.path", path]));
+    const isLocalRow =
+      fieldName.startsWith("_rowy_formulaValue_") ||
+      Boolean(find(tableRowsLocal, ["_rowy_ref.path", path]));
 
     const update: Partial<TableRow> = {};
 
@@ -422,6 +424,14 @@ export const updateFieldAtom = atom(
         row: localUpdate,
         deleteFields: deleteField ? [fieldName] : [],
       });
+
+      // TODO(han): Formula field persistence
+      // const config = find(tableColumnsOrdered, (c) => {
+      //  const [, key] = fieldName.split("_rowy_formulaValue_");
+      //  return c.key === key;
+      // });
+      // if(!config.persist) return;
+      if (fieldName.startsWith("_rowy_formulaValue")) return;
 
       // If it has no missingRequiredFields, also write to db
       // And write entire row to handle the case where it doesn’t exist in db yet
--- a/src/components/SideDrawer/MemoizedField.tsx
+++ b/src/components/SideDrawer/MemoizedField.tsx
@@ -6,6 +6,7 @@ import FieldWrapper from "./FieldWrapper";
 import { IFieldConfig } from "@src/components/fields/types";
 import { getFieldProp } from "@src/components/fields";
 import { ColumnConfig, TableRowRef } from "@src/types/table";
+import { TableRow } from "@src/types/table";
 
 export interface IMemoizedFieldProps {
   field: ColumnConfig;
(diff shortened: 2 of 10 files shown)


htuerker:
Hello, @mnmt7! 

First of all, thanks for the contribution!

Overall is quite good, I couldn't recognize any issue. I'll just ask some features here regarding my future concerns and give you some instructions how to implement with ease. 

First of all, problem is you're re-calculate formula function in SideDrawer. Which is good, if you don't make heavy computations or, most importantly dependent on deterministic code block. Whether the result is indeterministic or we have heavy computations(for example an http request), we don't want to re-calculate again.

Here's a simple example scenario: 
https://user-images.githubusercontent.com/46192266/234530411-763d18f2-082b-4375-93af-4a2f98951c92.mov

By now, we don't want to persist any calculation in firestore but later we may want to make it an option for our users. So, the trivial solution is to extend `omitRowyFields` 
```typescript
// src/utils/table
  delete shallowClonedRow["_rowy_ref"];
  delete shallowClonedRow["_rowy_outOfOrder"];
  delete shallowClonedRow["_rowy_missingRequiredFields"];
  delete shallowClonedRow["_rowy_new"];
  // !!
  delete shallowClonedRow["_rowy_formulaValue"];
```

Whenever the result has changed in useFormula, `useUpdateFieldAtom` to update `_rowy_formulaValue`, but make sure you don't persist data. Basically, just interrupt the execution when the fieldName is `_rowy_formulaValue`;

Also, you should figure out one more thing, there might be multiple formula fields in the row, so maybe you should consider using regexp to catch something like: `_rowy_formulaValue_columnKey`.

htuerker (changes requested):
First of all, problem is you're re-calculate formula function in SideDrawer. Which is good, if you don't make heavy computations or, most importantly, it's dependent on deterministic code block. Whether the result is indeterministic or we have heavy computations(for example an http request), we don't want to re-calculate again.

Here's a simple example scenario: 
https://user-images.githubusercontent.com/46192266/234530411-763d18f2-082b-4375-93af-4a2f98951c92.mov

By now, we don't want to persist any calculation in firestore but later we may want to make it an option for our users. So, the trivial solution is to extend `omitRowyFields` 
```typescript
// src/utils/table
  delete shallowClonedRow["_rowy_ref"];
  delete shallowClonedRow["_rowy_outOfOrder"];
  delete shallowClonedRow["_rowy_missingRequiredFields"];
  delete shallowClonedRow["_rowy_new"];
  // !!
  delete shallowClonedRow["_rowy_formulaValue"];
```

Whenever the result has changed in useFormula, `useUpdateFieldAtom` to update `_rowy_formulaValue`, but make sure you don't persist data. Basically, just interrupt the execution when the fieldName is `_rowy_formulaValue`. Then, use that value in SideDrawer instead of calculating. This will also help us to implement persistence support in future.

Also, you should figure out one more thing, there might be multiple formula fields in the row, so maybe you should consider using regexp to catch something like: `_rowy_formulaValue_columnKey`.

Please don't hesitate to reach out to me for further assistance. Feel free to put your own solution, I'd love to see a different solution in case you think anything better.

mnmt7:
@htuerker I have implemented the changes. Please review when possible.

src/atoms/tableScope/rowActions.ts
```diff
@@ -423,6 +425,14 @@ export const updateFieldAtom = atom(
...
       });
 
+      // TODO(han): Formula field persistence
+      // const config = find(tableColumnsOrdered, (c) => {
+      //  const [, key] = fieldName.split("_rowy_formulaValue_");
+      //  return c.key === key;
+      // });
+      // if(!config.persist) return;
+      if (fieldName.startsWith("_rowy_formulaValue")) return;
+
```
htuerker, line 435: Hey @mnmt7, I've just added it, it's required to prevent unnecessary db updates.

htuerker (approved):

htuerker:
Hey again, @mnmt7! Thanks for your contribution. LGTM 👍

Issue #1154, opened 2023-03-12
Expand support for fields that can be copied to clipboard


harinij:
Currently, we have a lot of field types, that can't be copy-pasted via the right-click context menu or hotkey.

Expand the coverage for the following fields to be copyable. For most fields, the user should be able to copy the field value of whatever is displayed in the cell.
- image/file cell should copy the URL to the clipboard
- single select, multiple select
- rating - value is copied
- slider
- date
- datetime
- duration
- lat/long

Also, check if any other fields can be added as well to the above list. Exceptions are programmable fields such as action, subtable, formula, derivatives, connector, formula which can't be copied at the moment

mnmt7:
@harinij I would like to work on this issue.

harinij:
@mnmt7 Thanks :) I have assigned it to you now.

PR #1233, opened 2023-04-30
Expand support for fields that can be copied to clipboard

mnmt7:
This change resolves the issue #1154.
Furthermore, the update includes a new functionality that limits the fields where values can be pasted using the hotkey. Prior to this update, the hotkey allowed for the pasting of values in any field type, including read-only fields like CreatedBy, CreatedAt. This created the possibility for data corruption, as read-only fields could be modified. The new feature prevents such accidental modifications by restricting the fields where the hotkey can be used, ensuring that read-only fields remain unmodifiable.


bofeiw (approved):
LGTM

PR #1248, opened 2023-05-07
Edit functionality for options of single select and multi select fields

### LINKED ISSUE #1247: Single and Multi select fields - edit for options
**Describe the issue**
Currently, the single and multi-select field options, do not have an editing option. They need to be deleted and added again if you want to edit them.

**Solution**
Add the option to click on an option value and edit the value of the option in this modal for single and multi-select

Screenshot 2023-05-06 at 1 36 07 pm

mnmt7:
Solves #1247

### DIFF
--- a/src/components/fields/SingleSelect/Settings.tsx
+++ b/src/components/fields/SingleSelect/Settings.tsx
@@ -14,6 +14,7 @@ import {
 } from "@mui/material";
 import AddIcon from "@mui/icons-material/AddCircle";
 import RemoveIcon from "@mui/icons-material/CancelRounded";
+import CheckIcon from "@mui/icons-material/CheckCircleRounded";
 import ColorSelect, {
   SelectColorThemeOptions,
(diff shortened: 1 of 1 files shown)


harinij:
@mnmt7 Thanks for submitting the PR. One quick update - can the UI be updated to also auto-save after an edit and clicked out of the text box? I.e someone should be able to double click on an option, edit it and click out of the cell to have the updated value saved instead of the current option to explicitly click the tick option on the right.
Screenshot 2023-05-24 at 12 19 29 pm

mnmt7:
@harinij I have added the functionality to auto-save when the user clicks outside the text field. 
Right now, the labels are editable when the user single-clicks on them. I just wanted to confirm whether it should be a single-click or double-click. Additionally, since the fields are now auto-saved when the user clicks outside the text field, should I remove the tick option or not?

harinij:
@mnmt7 Thanks, the updates look good as it is, we will get it merged soon.

Issue #1155, opened 2023-03-12
Right click menu on cell to have "Filter by" option to filter by that value of the cell


harinij:
The right-click context menu should have a `filter-by` option and then clicking this option applies a filter on the table → with the selected column and field cell value.

harinij:
> Hey @harinij, can I have this issue assigned to me?

Sure done ✅

PR #1251, opened 2023-05-09
"Filter by" option to filter by the value of the cell

mnmt7:
Solves #1155

### DIFF
--- a/src/components/Table/ContextMenu/MenuContents.tsx
+++ b/src/components/Table/ContextMenu/MenuContents.tsx
@@ -275,16 +275,6 @@ export default function MenuContents({ onClose }: IMenuContentsProps) {
         ? selectedColumn.config?.renderFieldType
         : selectedColumn?.type
     );
-    const handleFilterValue = () => {
-      openTableFiltersPopover({
-        defaultQuery: {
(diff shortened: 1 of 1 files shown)


harinij:
Hi @mnmt7 - Thank you for this PR. I tested this and it works well. Suggest a few minor changes.
1) The older "Filter value" option needs to be removed. That option used to open the filter popup modal and then user had to click Filter button again for the filter to show up. This newer "Filter by" option is the expected behavior to quickly filter by a value in a table. So the older "Filter value" be deprecated to avoid double options.
Screenshot 2023-05-24 at 12 37 28 pm

2) I also noticed that this "Filter by" option is disabled on fields which are spread out versions of a JSON - however manual filter on this field works well. It would be good if "Filter by" is enabled in this field as well as double checking for other field types. Quick [video explainer](https://share.descript.com/view/udG8px45VLp).
Screenshot 2023-05-24 at 12 35 17 pm

mnmt7:
@harinij, I have implemented the changes. Please review when possible.

htuerker (approved):
Hey, @mnmt7! Thanks for your contribution! LGTM!

Issue #1181, opened 2023-03-28
Safari Compatibility: Safari Number picker takes over and doesn't allow decimals.


BeeBombshell:
Originally posted by: HeikoK#9683 on Discord Support Forum

**Describe the bug**
 As soon as add data (currency) via my App, the data is correctly stored in Firebase and displayed in Rowy, but as soon as I click in the field, the values are set to zero. I figured out that this is a problem with Safari. Set aside that the table view is pushing the content out of the cell, the Safari Number picker takes over and only allows integers. Check also the editor to the right. Once you press either . or , the value is set to zero. I attached a small video showing this (The values are updated in Firebase and this is reflected in my mobile app).

**Screenshots**

https://user-images.githubusercontent.com/53828745/228250381-c5c01fff-d609-470f-a065-2ba66437c431.mp4

**Desktop (please complete the following information):**
 - OS: MacOS Ventura 13.2.1
 - Browser: Safari
 - Version: 16.3

zcesur:
hi @bofeiw, this is zaf from algora

for your command to go through, you need to sign up with your rowy.io email address (to join the [rowy org on algora](https://console.algora.io/org/rowyio)) & link your github account. sorry for the inconvenience!

bofeiw:
Thank @rjackson for fixing the issue and providing detailed explanation. Good job!

PR #1254, opened 2023-05-10
Fix Safari turning decimal number inputs to 0

rjackson:
Chrome has some slightly unique input[type="number"] handling behaviour, where it restricts the characters that can be typed to numeric characters. Under-the-hood, Chrome withholds certain invalid input states from being emitted as change events – so we only receive change events for valid numbers.

Safari behaves slightly differently. It allows any characters to be entered and emits a change event on each character press, but its internal "value" is only set if the typed input is a valid number. The change events then either come through with a numeric value represented, or as an empty string.

For example, when typing "12.34" we receive onChange events with "1", "12", "", "12.3", and "12.34".

On that third onChange event when we pass "" up to React, React happily ignores the change (I _think_ bceause the incoming value "" already matches the element's value of ""; nonetheless, React have solved this issue for us).

When we parse the input via `Number(v)` we encounter problems. `Number("")` resolves to `0`, React sets the input's value to `0`, and the user is therefore unable to type decimal values successfully.

The solution for this is not to cast to a number at all. We'll rely on default browser behaviour to manage the input for us (allowing users to enter invalid characters), but we're still safe because the element's internal state will always be numeric.

In the scenario that a customer does try enter an invalid input "abc123", the input field simply blanks itself when they leave it.

/claim #1181

### DIFF
--- a/src/components/Table/TableCell/EditorCellTextField.tsx
+++ b/src/components/Table/TableCell/EditorCellTextField.tsx
@@ -4,6 +4,7 @@ import { spreadSx } from "@src/utils/ui";
 
 export interface IEditorCellTextFieldProps extends IEditorCellProps<string> {
   InputProps?: Partial<InputBaseProps>;
+  onBlur?: () => void;
 }
 
 export default function EditorCellTextField({
@@ -11,6 +12,7 @@ export default function EditorCellTextField({
   value,
   onDirty,
   onChange,
+  onBlur,
   setFocusInsideCell,
   InputProps = {},
 }: IEditorCellTextFieldProps) {
@@ -19,7 +21,12 @@ export default function EditorCellTextField({
   return (
     <InputBase
       value={value}
-      onBlur={() => onDirty()}
+      onBlur={() => {
+        if (onBlur) {
+          onBlur();
+        }
+        onDirty();
+      }}
       onChange={(e) => onChange(e.target.value)}
       fullWidth
       autoFocus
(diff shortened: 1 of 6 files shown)


bofeiw (changes requested):
Hi @rjackson, nice fix. It worked in the table editor cell. 

1. Can you apply the change to the side panel editor as well? (the right side panel when you click the expand button on the right)
2. Noticed that invalid input in Safari (e.g. "12,3") will be saved as empty string instead of 0. Can this be fixed?
3. Noticed that input "e" will reset the number to 0. It might be a conflict with scientific number (e.g. 1e9). Out of scope for this PR and there may not be a simple solution. Do not need fix in this PR, however if you have an idea welcome to contribute.

rjackson:
@bofeiw Apologies, I hadn't spotted the side panel editor!

I've got the PR updated with those changes now. The SideDrawerField works a little differently to EditorCell, so there's a _slight_ difference between the two pieces of code, but the overall pattern is the same.

Whilst a user is typing (onChange), we're acting upon any non-invalid inputs (skipping over empty strings). I've re-added the `onChange` code blocks even though they're _technically_ unnecessary here, but it lets me add a comment explaining what's going on.

Once a user has finished typing and leaves the field (onBlur), we're then casting the values to Number – in case they weren't already. This ensures the value we store in Firebase is number – either 0, or a valid number.

The EditorCell also supports completing the change via the Enter key, which was still saving empty strings to Firebase. To solve that I've adjusted the Enter-key code to invoke `onBlur` if we've given it one. As far as I can tell, that now guarantees all inputs will be cast to Number before being saved.

Finally, I've applied the same fixes to the Percentage field type. There are a few other field types that are represented by numbers under-the-hood, but Number and Percentage were the two allowing input via `input[type=number]` and therefore susceptible to this issue. The other number fields (Slider, Rating) have custom inputs that aren't affected by this bug.

bofeiw (approved):

harinij:
Thanks @rjackson for the contribution 🎉

PR #1255, opened 2023-05-10
Add colour change customisation to Slider

rjackson:
As requested in #519, but hasn't had any progress made since December.

This PR implements the same colour scale configuration as used by the Percentage component, allowing users to change the start, mid, and end colours used in the little Slider visualisation:

In my coming example screenshot, I have 3 sliders configured:

1. Set up before colour support was present, proving fallback to default colours when none are configured
2. Customised colours
3. Default colours, saved against column during initial configuration

Screenshots:

Description | Screenshot
---- | ----
Example with regular and customised sliders | Screenshot 2023-05-10 at 20 41 47
Default settings | Screenshot 2023-05-10 at 20 42 00
Customised colour settings | Screenshot 2023-05-10 at 20 42 10

/claim #519


harinij:
Hi @rjackson, reviewed and tested the PR now and it works well. One minor change is required to support the dark mode. Currently in the dark mode the color picker looks as follows:

Screenshot 2023-05-30 at 11 24 20 am

However, it should look like this (similar to the color picker implementation in the `Percentage` field)
Screenshot 2023-05-30 at 11 24 15 am

You can change between light and dark theme from the profile section
image

rjackson:
Ah, thanks @harinij 

That looks to have been caused by me creating my branch off of `main`, where `ColorPickerInput` doesn't yet have dark mode support (added in 0ff87300b46cfc74d7a794b8d78bae8abb4d178a).

I've rebased my work on top of the latest `develop` branch and all seems correct now

Issue #1071, opened 2023-01-09
Unique Rowy URLs per filter combination


harinij:
### Discussed in https://github.com/rowyio/rowy/discussions/1065

<div type='discussions-op-text'>

<sup>Originally posted by **gregdingle** January  8, 2023</sup>
It would be very useful to get a URL that represents the current filtered view of a table.
1) share with colleagues
2) bookmark for self
3) link to from other tables as a way of show foreign key data (id in one table cell links to filtered view in another table)

For example:

https://rowy.app/p/test/table/documents#filters:documentData.copy_count>0

<img width="890" alt="CleanShot 2023-01-07 at 08 15 00@2x" src="https://user-images.githubusercontent.com/28797/211152637-69fe14ea-c95e-4dd7-af61-c112bbd6890d.png">
</div>

staticGuru:
@harinij , I am working on the changes

harinij:
Hi @staticGuru Great thanks, I have assigned it to you. Join our Discord community and feel free to chat on the #contribution channel in case you need any info.

staticGuru:
Hi @harinij 
I hope you doing well.
I worked on the features and created PR #1256 .
Kindly check

PR #1256, opened 2023-05-11
Filter url

staticGuru:
#1071 
- Added new feature: filter parameters in page URL.
- Ability to edit filter parameters using URL elements.
- Included loom link for reference. (https://www.loom.com/share/ddf8a5b2817a4700a518054e1b5d7d4e)
- Pull request submission for Issue #1071.

### DIFF
--- a/src/types/table.d.ts
+++ b/src/types/table.d.ts
@@ -184,7 +184,8 @@ export type TableFilter = {
     | "time-minute-equal"
     | "id-equal"
     | "color-equal"
-    | "color-not-equal";
+    | "color-not-equal"
+    | "-is-";
   value: any;
 };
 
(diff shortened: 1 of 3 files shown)


staticGuru:
@harinij Are there any updates regarding this PR?

src/types/table.d.ts
```diff
@@ -184,7 +184,8 @@ export type TableFilter = {
     | "time-minute-equal"
     | "id-equal"
     | "color-equal"
-    | "color-not-equal";
+    | "color-not-equal"
+    | "-is-";
```
shamsmosowi, line 188: hi @staticGuru thanks for the contribution, everything looks good!, i don't understand the difference between `id-equal` and `-is-`

staticGuru: @shamsmosowi Actually, I add the `-is-` keyword to make the similarities between URL values and filter UI values.

shamsmosowi: I tried removing it and use `id-quals` directly, and it seems to work fine, did you mainly add it for URL readability, or are there other needs for it?

staticGuru: @shamsmosowi No, Actually `-is-` is used to get the operators and operants values from the given URL sections.
if you use the `id-equals` in the URL sections, it will affect the URL readability and also we can't separate the operators and operants values from the URL.

PR #1268, opened 2023-05-23
Sort button to showcase if a column is sorted or not

### LINKED ISSUE #1244: UI Improvement to showcase if a column is sorted or not
**Background**
Currently, when sorting is applied on a column there is no visual indicator to show which column the sorting is applied on. This results in confusion for users.

**Solution **
Add a button similar to the Filter button for indicating whether sorting is applied on a table or not. 
Screenshot 2023-05-04 at 3 26 54 pm

Also when sorting is applied it shows the column by which the sorting is applied and the order.
Screenshot 2023-05-04 at 3 28 15 pm

mnmt7:
closes #1244 

This PR adds a sort button which is visible only when the sort is applied on a column. Clicking on sort button opens a popover where the user can select other columns for sorting and as well as select the sorting direction i.e. ascending or descending.

Also on a side note, I noticed that even if you remove the sort, it is not removed from the Firebase and if you reload the page after removing the sort, it will be reapplied on the table. This should be an easy fix and I can do it in this PR if you say so.

/claim #1244

### DIFF
--- a/src/components/ColumnMenu/ColumnMenu.tsx
+++ b/src/components/ColumnMenu/ColumnMenu.tsx
@@ -192,9 +192,10 @@ export default function ColumnMenu({
         setTableSorts(
           isSorted && !isAsc ? [] : [{ key: sortKey, direction: "desc" }]
         );
-        if (!isSorted || isAsc) {
-          triggerSaveTableSorts([{ key: sortKey, direction: "desc" }]);
-        }
+
+        triggerSaveTableSorts(
+          isSorted && !isAsc ? [] : [{ key: sortKey, direction: "desc" }]
+        );
         handleClose();
       },
       active: isSorted && !isAsc,
@@ -209,9 +210,9 @@ export default function ColumnMenu({
         setTableSorts(
           isSorted && isAsc ? [] : [{ key: sortKey, direction: "asc" }]
         );
-        if (!isSorted || !isAsc) {
-          triggerSaveTableSorts([{ key: sortKey, direction: "asc" }]);
-        }
+        triggerSaveTableSorts(
+          isSorted && isAsc ? [] : [{ key: sortKey, direction: "asc" }]
+        );
         handleClose();
       },
       active: isSorted && isAsc,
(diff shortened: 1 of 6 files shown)


iamanishroy:
> Also on a side note, I noticed that even if you remove the sort, it is not removed from the Firebase and if you reload the page after removing the sort, it will be reapplied on the table. This should be an easy fix and I can do it in this PR if you say so.

Hey @mnmt7, it would be great if you could work on it. Also, please check the requested change as it is related to this fix.

mnmt7:
Hey @iamanishroy, didn't get what you meant by the requested change?

harinij:
Hi @mnmt7 I think @iamanishroy is indicating that it would be great if you can also work on the following point you raised as part of this PR.

> Also on a side note, I noticed that even if you remove the sort, it is not removed from the Firebase and if you reload the page after removing the sort, it will be reapplied on the table. This should be an easy fix and I can do it in this PR if you say so.

iamanishroy (changes requested):

mnmt7:
@iamanishroy, I have implemented the requested changes Now, when a user removes the sort and reloads the page, the previous sort won't be reapplied. As per your suggestion, I have utilized the `triggerSaveTableSorts()` function to achieve this. However, I would like to confirm one thing. It seems to me that `triggerSaveTableSorts()` is intended to modify sort settings for individual users rather than applying them for all users. Please clarify if my understanding is correct.

The current flow of the sort is as follows:

1. The user applies the sort.
2. A popup appears, giving the option to save the sort for all users.
3. If the user chooses not to save the sort and:
    - The user removes the sort and refreshes the page, the sort won't be reapplied.
4. If the user chooses to save the sort and:
    - The user removes the sort, the changes are immediately reflected on the screen i.e. the sort is removed. However, if the page is reloaded, the sort will be reapplied because now the "sort settings for all users" will take the precedence. In short, if no user-specific sort setting exists, the sort settings for all users will be utilized.

Furthermore, there was one more thing I wanted to mention. Regarding the design of the columns dropdown, I made a slight modification. Instead of the design shown in the provided images, I have opted for a slightly different design:

Requested design -
236119112-caed044e-1c03-4c25-b054-1428d97ff38d

Implemented design -
Screenshot 2023-06-01 at 10-16-30 Rowy – GCP as easy as ABC

I chose this design since the dropdown component was already being used for the filter functionality, and it seemed logical to maintain consistency. The only issue with this was what should be the action performed when the "clear" option is clicked. Right now, the sort is removed and consequently the sort modal disappears.

iamanishroy (approved):
LGTM

PR #1279, opened 2023-06-06
chore: migrate from CRA and Jest to Vite and Vitest

### LINKED ISSUE #1275: Chore: migrate to Vite
Setup the project with Vite

il3ven:
This PR migrates the codebase from CRA and Jest to Vite and Vitest.

Changes:
- Update public folder in `firebase.json` to `dist` from `build`
- Change `REACT_APP` env variables to `VITE_APP`
- Move `index.html` to root as it is required by Vite
- Update packages like `use-debounce`, `json2csv` as their newer versions have proper support for ESM
- Remove `esModuleInterop` from `tsconfig` to keep our app as close as possible to create-vite-app template
- Instead of Webpack's `raw-loader` use Vite specific loader argument
- Use `import.meta.env` instead of `process.env` as recommended by Vite
- Migrated to Vitest from Jest. Vitest has better ecosystem integration with Vite and its API is also similar to Jest.
- Add colord because it was missing from package.json.

Checking out to this PR won't work. To test locally you'll have to checkout to rowyio/form-builder#10 and use `yarn link`. I have made some changes to that package. Reason for the change is explained in rowyio/form-builder#10.

I have tested the changes locally after linking to @rowy/form-builder. `yarn build`, `yarn start` work as expected. `yarn test` fails with a firebase `auth/network-request-failed`. _( Help needed )_

For this PR, I have used [@vitejs/plugin-react](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react) and not [@vitejs/plugin-react-swc](https://github.com/vitejs/vite-plugin-react-swc) because jotai doesn't support fully support swc. I did try swc at first but it didn't work.

Fixes #1275 
/claim #1275


shamsmosowi:
Thank you @il3ven ! it works perfectly!

PR #1291, opened 2023-06-11
add support for bulk deletion of rows

### LINKED ISSUE #809: Support bulk deletion of rows
Currently, you can only delete a row one at a time. Need ability to select and delete multiple rows at a time.

Design file: [Link](https://www.figma.com/file/2ajhGblHmRYOiQSv0LHcyN/Multi-select?type=design&node-id=0-1&t=IBwEHcMDSCxl5MRO-0)

**Note**
- Implement only the table part of the design and not any other elements.
- When rows are selected the delete button options shows up on top of the page
- In case of any questions feel free to ask here or on [Discord](rowy.io/discord) to ensure you have good clarity on the design and implementation needed

il3ven:
This PR adds support for bulk deletion of rows by implementing a checkbox next to each row. The implementation is based on the [example](https://tanstack.com/table/v8/docs/examples/react/row-selection) provided by react-table.

#### Needs more attention
- Selecting a few rows should put the header checkbox in an [indeterminate](https://mui.com/material-ui/react-checkbox/#indeterminate) state. For some reason it is not working. I don't see any problem with my code/logic.
- Sometimes, after deleting rows the [`<LoadedRowStatus />`](https://github.com/rowyio/rowy/blob/develop/src/components/TableToolbar/LoadedRowsStatus.tsx) component doesn't update the total number of rows. I request someone with more experience of the codebase to look into it.
- The state for managing selected rows is set by react-table. They [set the ids](https://github.com/il3ven/rowy/blob/9b84822a1c31ac9cc5b08609f33b1dcbaedbf779/src/components/Table/Table.tsx#L234) of selected rows into the state. The [id is equal to `_rowy_ref.path`](https://github.com/il3ven/rowy/blob/9b84822a1c31ac9cc5b08609f33b1dcbaedbf779/src/components/Table/Table.tsx#L61) and the atom delete [function accepts a path](https://github.com/il3ven/rowy/blob/2715a8094aa740bd6c783ad7e28fb24f6973cded/src/atoms/tableScope/rowActions.ts#L222). In case the id is not equal to `_rowy_ref.path` the implementation will fail. I think it is not a point of worry because id should always be equal to `_rowy_ref.path`. Please correct me if I am wrong.

Note: Check the review comments for context about the decisions I have made.

/claim #809
Closes #809

https://github.com/rowyio/rowy/assets/4337699/54369216-fc09-40dc-b601-b41fc7873163

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -276,11 +276,13 @@ export default function Table({
 
       updateColumn({
         key: result.draggableId,
-        index: result.destination.index,
+        index: selectedRows
+          ? result.destination.index - 1
+          : result.destination.index,
         config: {},
       });
     },
-    [updateColumn]
+    [updateColumn, selectedRows]
   );
 
   const fetchMoreOnBottomReached = useThrottledCallback(
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -34,6 +34,7 @@ import {
   selectedCellAtom,
   tableSortsAtom,
   tableIdAtom,
+  serverDocCountAtom,
 } from "@src/atoms/tableScope";
 import { projectScope, userSettingsAtom } from "@src/atoms/projectScope";
 import { getFieldType, getFieldProp } from "@src/components/fields";
@@ -112,6 +113,7 @@ export default function Table({
   selectedRows,
 }: ITableProps) {
   const [tableSchema] = useAtom(tableSchemaAtom, tableScope);
+  const [serverDocCount] = useAtom(serverDocCountAtom, tableScope);
   const [tableColumnsOrdered] = useAtom(tableColumnsOrderedAtom, tableScope);
   const [tableRows] = useAtom(tableRowsAtom, tableScope);
   const [tableNextPage] = useAtom(tableNextPageAtom, tableScope);
@@ -159,24 +161,39 @@ export default function Table({
       );
     }
 
-    if (selectedRows)
-      _columns.unshift(
+    return _columns;
+  }, [tableColumnsOrdered, canAddColumns, canEditCells, selectedRows]);
+
+  columns.unshift(
+    ...useMemo(() => {
+      if (!selectedRows) return [];
+
+      return [
         columnHelper.display({
           id: "_rowy_select",
           size: 41.8, // TODO: We shouldn't have to change this often
-          header: ({ table }) => (
-            <FormControlLabel
-              sx={{ margin: 0 }}
-              label=""
-              control={
-                <Checkbox
-                  checked={table.getIsAllRowsSelected()}
-                  indeterminate={table.getIsSomeRowsSelected()}
-                  onChange={table.getToggleAllRowsSelectedHandler()}
-                />
-              }
-            />
-          ),
+          header: ({ table }) => {
+            const checked =
+              Object.keys(selectedRows.state).length >= serverDocCount!;
+            const indeterminate = Object.keys(selectedRows.state).length > 0;
+            return (
+              <FormControlLabel
+                sx={{ margin: 0 }}
+                label=""
+                control={
+                  <Checkbox
+                    checked={checked}
+                    indeterminate={indeterminate && !checked}
+                    onChange={() => {
+                      table.toggleAllRowsSelected(
+                        !table.getIsAllRowsSelected()
+                      );
+                    }}
+                  />
+                }
+              />
+            );
+          },
           cell: ({ row }) => {
             return (
               <FormControlLabel
@@ -192,11 +209,10 @@ export default function Table({
               />
             );
           },
-        })
-      );
-
-    return _columns;
-  }, [tableColumnsOrdered, canAddColumns, canEditCells, selectedRows]);
+        }),
+      ];
+    }, [selectedRows])
+  );
 
   // Get user’s hidden columns from props and memoize into a `VisibilityState`
   const columnVisibility: VisibilityState = useMemo(() => {
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -3,6 +3,7 @@ import { useMemo, useRef, useState, useEffect, useCallback } from "react";
 import { useAtom, useSetAtom } from "jotai";
 import { useThrottledCallback } from "use-debounce";
 import {
+  RowSelectionState,
   createColumnHelper,
   getCoreRowModel,
   useReactTable,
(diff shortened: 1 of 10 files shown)


src/components/Table/Table.tsx
```diff
@@ -75,6 +77,20 @@ export interface ITableProps {
...
+   * If defined, it will show a checkbox to select rows. The
+   * state is to be maintained by the parent component.
+   *
+   * Usage:
+   *
+   * const [selectedRows, setSelectedRows] = useState<RowSelectionState>({});
+   * const selectedRowsProp = useMemo(() => ({state: selectedRows, setState: setSelectedRows}), [selectedRows, setSelectedRows])
+   * <Table selectedRows={selectedRowsProp} />
+   */
+  selectedRows?: {
```
il3ven, line 90: I have added this prop because we need the state outside of `<Table />` so that it can be passed to `<TableToolbar />`. The other option was to pull the react-table upto `<TablePage />`.

src/components/Table/TableBody.tsx
```diff
@@ -48,7 +54,7 @@ export interface ITableBodyProps {
  * - Renders row out of order indicator
  * - Renders next page loading UI (`RowsSkeleton`)
  */
-export const TableBody = memo(function TableBody({
```
il3ven, line 51: I had to remove memo because with it the component wasn't updating on row selection. I think it is because on row selection the `table` prop changes but memo does a shallow rendering and thus doesn't re-render the component.

src/components/Table/TableHeader.tsx
```diff
@@ -34,13 +35,14 @@ export interface ITableHeaderProps {
  *
  * - Renders drag & drop components
  */
-export const TableHeader = memo(function TableHeader({
-  headerGroups,
+export const TableHeader = function TableHeader({
+  table,
```
il3ven, line 39: I had to change `headerGroups` to `table` because `headerGroups` don't update on row selection.

iamanishroy: Hey @il3ven , I tried using `headerGroups` and keeping the memo on `TableHeader` and `TableBody`, and it's working fine for me. Could you please check it and let me know if I'm missing anything? Thank you!

il3ven: It is working right now with `headerGroups` and `memo`. It doesn't make sense at the moment but it is working. We can keep `memo` and `headerGroups`.

iamanishroy: Great! So, can you make the changes, as now it is working with `headerGroups` and `memo`.

il3ven: > So, can you make the changes, as now it is working with headerGroups and memo.

@iamanishroy Done!

iamanishroy:
Hey @il3ven, we also have a field type called Array-subtable. Although the view appears as a table, the internal workings are a bit different. The entire array subtable refers to a single field in a document, which is an array of maps. This is an edge case that is not handled here.

https://github.com/rowyio/rowy/assets/62830866/e2f1a8e0-c320-4aff-8601-c152640e90bc

harinij:
@il3ven tested this and it works well overall. There is just one slight issue, when you `freeze` any column then causes a UI issue as seen in the video below.

https://github.com/rowyio/rowy/assets/307298/181c57ca-f8a8-425d-ad9c-5888a8a17681

il3ven:
I'll try to fix all these issue by Friday. Sorry for the delay, really busy with other stuff.

il3ven:
I have pushed three commits addressing the comments.

### https://github.com/rowyio/rowy/pull/1291/commits/eef15acf0aaa76fa97cb0a0e7de77f39002be0ae

This commit fixes the freeze issue reported by @harinij.

### https://github.com/rowyio/rowy/pull/1291/commits/8d81c36b4c0bffba61b6f1524aeca1d239120c4a

I found a new bug. The reordering of rows was not working as expected. Fixed it.

### https://github.com/rowyio/rowy/pull/1291/commits/dc972d7346ff5ec2930abed6b646017ceae834e9

This commit addresses the [comment](https://github.com/rowyio/rowy/pull/1291#issuecomment-1600330500) by @iamanishroy. The solution for now is to disable row selection for array sub table.

Before explaining the reason for disabling row selection let me explain the cause for the reported bug.

#### Cause for the array sub table bug

We have the following logic for generating row ID.

```ts
const getRowId = (row: TableRow) => row._rowy_ref.path || row._rowy_ref.id;
```

All rows in case of array sub table have the same `_rowy_ref.path` and hence the same row ID. Therefore, causing the bug.

#### Giving unique row ID in case of array sub table

We can have the following logic to make row IDs unique.

```ts
const getRowId = (row: TableRow) =>
  row._rowy_ref.path + row._rowy_ref.arrayTableData?.index
    ? `.${row._rowy_ref.arrayTableData?.index}`
    : "";
```

This fixes the bug shown in the video [above](https://github.com/rowyio/rowy/pull/1291#issuecomment-1600330500). However, there is another bug that is the bulk delete functionality doesn't work. I have explained it below.

#### Why can't we bulk delete rows in case of array sub table?

Suppose we have three items in our array that is `[a, b, c]` and user has selected row index `1` and `2` to be deleted. The current delete function accepts only one index. Therefore, we'll have to delete sequentially. In code, it will look like following.

```ts
const [ _deleteRowDb ] = useAtom(_deleteRowDbAtom)

// This works as expected. The new array in DB is [a, c].
_deleteRowDb(path, {index: 1})  

// This does not work. As there is no item at index 2.
_deleteRowDb(path, {index: 2})  
```

The above explanation is simplified for the purpose of this explanation.

##### Solutions for bulk delete in case of array sub table

The best solution would be to ditch array sub table. The sub table should be a collection. We'll also be able to reuse existing code.

The other solution can be to assign unique IDs to each item in the array. The ID can be autogenerated. Instead of deleting item by index we can center our logic to delete item by ID.

Due to the above problem in deletion of rows. I feel it is best to disable row selection for now and implement a fix in a separate PR.

harinij:
Hi @il3ven thanks for the detailed explanation and updates. It appears that this PR has build errors, screenshot below - that would need further attention.
Screenshot 2023-07-18 at 4 36 03 pm

il3ven:
@harinij fixed the build error.

il3ven:
@shamsmosowi I have fixed two issues. Kindly review and merge.
- Show indeterminate state on selecting all rows if there are more rows in DB but they haven't been loaded
- Indeterminate state not working. This was due to a problem in the icon's CSS.

https://github.com/rowyio/rowy/assets/4337699/8a2aadec-6efd-4a56-b954-20a38af91075

Issue #1284, opened 2023-06-08
Import subcollection data/ auto create subtable & array-subtable schemas


shamsmosowi:
Currently, it's only possible to import collection data and auto generate schema for primary tables

Rowy should be able to detect and suggest the schema for both subtables and array subtables

The experience should be the same as the importing data/schema for primary collections

rishipurwar1:
@shamsmosowi working on this issue!

rishipurwar1:
> Rowy should be able to detect and suggest the schema for both subtables and array subtables

Hello @shamsmosowi I did some research on how we're fetching the data from firestore for each collection:
1. In `TableSourceFirestore.js`, we call the `useFirestoreCollectionAtom` hook to fetch the docs from firestore and save them to `tableRowsDbAtom`.
2. Then, we access the `tableRowsDbAtom` in the `table.ts` file, which contains all the documents for a particular collection.
3. Then we merge the two arrays after removing the duplicates and update the value of `tableRowsAtom`.
4.  Finally, we access the `tableRowsAtom` in the `EmptyTable` component to show the user that you already have some data in the firestore.

Am I right up to this point?

Now, to retrieve data for subcollections, we can use the same approach as we do for collections since fetching data for collections (`collection(firebaseDb, pathToCollection)`) and subcollections (`collection(firebaseDb, pathToSubCollection)`) works in the same way. However, we need to provide a way for the user to input the names of the subcollections, as each document can have multiple subcollections.

Approaches:
1. Provide a new field to the user at the time of filling out the "Create table" form just below the "Collection name" field.
2. Add one more tab in the "import" modal component to enter the subcollection names.
<img width="519" alt="Screenshot 2023-06-13 at 9 50 28 AM" src="https://github.com/rowyio/rowy/assets/57035408/711921f6-eaae-47a1-8641-07aea8447dd7">

Another consideration is how we should present the subcollection data. For example, should we generate separate table pages for each subcollection, and how will users access these specific subcollection tables?

And Could you please tell me what you mean by "subtables" and "array subtables"?

Please let me know if you have any further thoughts or if there are any other aspects we should consider regarding the display and access of subcollection data. Thank you!

shamsmosowi:
> > > Rowy should be able to detect and suggest the schema for both subtables and array subtables
> > 
> > 
> > Hello @shamsmosowi I did some research on how we're fetching the data from firestore for each collection:
> > 
[...]

PR #1302, opened 2023-06-17
Implement auto import for array subtables

rishipurwar1:
This PR fixes issue #1284. Now, Rowy can detect and suggest the schema for the array suitable.

https://github.com/rowyio/rowy/assets/57035408/327b92aa-dda0-4e1b-b950-f11b7b3de59c

### DIFF
--- a/src/components/Table/EmptyTable.tsx
+++ b/src/components/Table/EmptyTable.tsx
@@ -33,7 +33,7 @@ export default function EmptyTable() {
       : false;
   let contents = <></>;
 
-  if (tableSettings.isCollection !== false && hasData) {
+  if (hasData) {
     contents = (
       <>
         <div>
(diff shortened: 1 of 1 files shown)


rishipurwar1:
@harinij @shamsmosowi please review my PR.

src/components/Table/EmptyTable.tsx
```diff
@@ -33,17 +33,27 @@ export default function EmptyTable() {
       : false;
   let contents = <></>;
 
-  if (tableSettings.isCollection !== false && hasData) {
+  if (
+    (tableSettings.isCollection !== false ||
+      tableSettings.subTableKey?.length) &&
```
shamsmosowi, line 38: @rishipurwar1 how is `tableSettings.subTableKey?.length` used here, could we just remove `tableSettings.isCollection !== false` part of the condition?

rishipurwar1: > could we just remove `tableSettings.isCollection !== false` part of the condition?

Yes, we can do that. That way, we only need to check for the `hasData` flag, if it's true then show the import existing data template.

rishipurwar1: Updated the if condition.

shamsmosowi: @iamanishroy what do you think? is there any other conditions we should watch out for?

iamanishroy: LGTM 👍

iamanishroy (approved):
Seems fine 👍

PR #1306, opened 2023-06-19
Add code editor quick action to get secrets

bofeiw:
This PR adds support for code editor quick replacements for `rowy.secrets.get` method if the provided secret name does not exist.

The list of replacement values are sorted by similarity of the provided secret name. For example, if user types a typo "whatapp" and the correct secret "whatsapp" exists in GCP Secret Manager, it will appear on the top of the list.

Screenshot 2023-06-19 at 23 24 58

Screenshot 2023-06-19 at 23 25 16


il3ven (approved):
Approving the PR. 👍🏽

However, a note for documentation purposes. The quick fix will not be shown if the user has created only one secret. This is because in that case the error message generated by typescript is different and does not contain the word "SecretNames".

I have also pushed a commit fixing a minor syntax problem which is not related to this PR but I saw it while reviewing.

Issue #533, opened 2021-10-01
Drag and Drop fill for Text, Numbers and other fields 


harinij:
**Is your feature request related to a problem? Please describe.**
Drag and drop ability to copy values from one cell to another similar to the feature in excel or google docs. React data grid supports this feature. Enable and make this option workable such that it is copied over from one cell to rest of the cells below that it is dragged to. Previously we had this option however it is alpha version and has been disabled. Please review to make this option work again

Ali7040:
I want to work on it. Please assign this issue to me.

ashutosh887:
@harinij @BeeBombshell  Let me attempt this one if not resolved yet!

mithson:
@harinij is problem fixed? if it still persists I can give it a try just let me know so I can work on this one

siddhakdak:
is this issue already solved or not ? I want to work on this

siddhakdak:
@Ali7040 had you solved this issue or not

harshlakhara:
> **Is your feature request related to a problem? Please describe.** Drag and drop ability to copy values from one cell to another similar to the feature in excel or google docs. React Table supports this feature. Enable and make this option workable such that it is copied over from one cell to rest of the cells below that it is dragged to. Previously we had this option however it is alpha version and has been disabled. Please review to make this option work again

@harinij just want a clarifaction that 
if i drag some value from the cell I shoulld be able to paste the value in the same column cells which have been dragged over?

PR #1346, opened 2023-07-18
Fix#533

Ali7040:
Copy & Paste issue in cells solved

/claim #533


harinij:
@Ali7040 can you share a bit about how this change does the drag and drop fill change with a screenrecording it would be useful. Also if you can remove any unnecessary comments in the code added it would be good.

Ali7040:
> @Ali7040 can you share a bit about how this change does the drag and drop fill change with a screen recording it would be useful. Also if you can remove any unnecessary comments in the code added it would be good.

Sure!

Issue #1347, opened 2023-07-19
Bug: Offline error


harinij:
**Describe the bug**
We are seeing occasional cases where user gets "You're offline" error and after that, they get stuck in that state, even though they are online and have internet connectivity.

This is a bug we have not been able to actively reproduce but have a bunch of users facing and complaining about [here](https://discord.com/channels/853498675484819476/1125514154489425940).

Previously this issue was seen [here](https://github.com/rowyio/rowy/issues/1289#issue-1751327460) and resolved. However, this need further investigation to resolve the issue completely.

**Expected behavior**
Check the logic where "You're offline" comes into play and debug if there is any scenario where it is falsely triggered. Remove this logic from places where it is not required. This error should not come in when the user is online and the state should not be stuck there.

**Screenshots**
![image](https://github.com/rowyio/rowy/assets/307298/c27c6cbd-a432-4d9e-b7cf-6e331f2a1dcd)

**Additional context**
Feel free to discuss any technical details with the team on [discord](rowy.io/discord) if you want to chat for further clarity as this is an ambitious bug.

hrutik7:
Hey ,
can you assign this to me ?

vats147:
@harinij 
Can we make 5 second timer whenever user again come to online then we automatically reload page after 5 second?
It will decrease with time.

feliciien:
can I work on this?

hrutik7:
Just for the context 
The first priority here is reproduce this bug in local dev .
anyone who want to contribute on this specific issue you can follow the support channel discussion that will be really helpful :)

dylanrod21:
@feliciien I've added some additional details in the comments on the pull request here https://github.com/rowyio/rowy/pull/1352

ashutosh887:
@harinij @BeeBombshell  Let me attempt this one if not resolved yet!

ashutosh887:
@harinij @BeeBombshell Please update me on this so I'll proceed

harinij:
Hi @ashutosh887 we have an open PR here https://github.com/rowyio/rowy/pull/1352 however since we have not be able to replicate this fully in local env and the issue seems to happen to only a small set of users we are still having this issue open as there is no complete solution yet.

@vats147 and @dylanrod21 thanks for investigating this and your contribution, we are currently reviewing the PR and will post comments on it soon.

Anurag00321:
[...]

PR #1352, opened 2023-07-24
Solve Offline Error

vats147:
Solve #1347 
/claim #1347
We are seeing occasional cases where user gets "You're offline" error and after that, they get stuck in that state, even though they are online and have internet connectivity.
image

So if user come online then i add 5 second timer after that page will automatically reload.
image

### DIFF
--- a/src/components/ErrorFallback.tsx
+++ b/src/components/ErrorFallback.tsx
@@ -21,6 +21,25 @@ export function ErrorFallbackContents({
   ...props
 }: IErrorFallbackProps) {
   const isOffline = useOffline();
+  const [isOnline, setIsOnline] = useState(window.navigator.onLine);
+
+  useEffect(() => {
+    function handleOnlineStatus() {
+      setIsOnline(true);
+    }
+
+    function handleOfflineStatus() {
+      setIsOnline(false);
+    }
+
+    window.addEventListener("online", handleOnlineStatus);
+    window.addEventListener("offline", handleOfflineStatus);
+
+    return () => {
+      window.removeEventListener("online", handleOnlineStatus);
+      window.removeEventListener("offline", handleOfflineStatus);
+    };
+  }, []);
 
   if ((error as any).code === "permission-denied")
     return (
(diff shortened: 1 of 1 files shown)


dylanrod21:
@vats147 Were you able to reproduce this bug at all locally, based on the testing I had done when looking into this issue I'm not sure that this change would resolve the error. I currently in production have not been able to use the service at all, any account I try to sign up with is bringing me right to the 'You're Offline' message with the following errors in the console. In my local build of the app I have no issues using the service. I'm curious to see if anyone has been able to locally reproduce the issue I'm seeing in prod to better diagnose what may be the root cause of the bug.
image

vats147:
@dylanrod21 Now the issue has been solve kindly check it

vats147:
@dylanrod21 Can you share Screenshot where is error occur ?

dylanrod21:
@vats147  Here's a video, I'm off to bed now, let me know if this helps.
[Rowy Offline Error_ Jul 26,2023 00_40 AM.webm](https://github.com/rowyio/rowy/assets/6277685/f5b74feb-ab83-428b-81e8-9a8547fe66a5)

hrutik7:
hey @vats147 @dylanrod21 
any luck to reproduce this bug on local ?

vats147:
@hrutik7 No i don't find any clue

g1mishra:
[> webm](https://github.com/rowyio/rowy/pull/1352#issuecomment-1650971192)

The video clearly shows a problem with the auth/token API, which is causing the issue we're facing. 
The ErrorBoundary's Fallback Component is responsible for displaying the `"You’re offline"` message, determined by whether `error.message.includes("Failed to fetch")` evaluates to true.

However, it's important to understand that "Failed to fetch" doesn't always mean you're offline; there could be other reasons. 

To fix this problem, we need to check the logs to get more information. Unfortunately, we can't reproduce the error locally because it's happening on the API side.

il3ven:
This PR doesn't solve the offline issue. The issue is not on the frontend side. The users affected with the offline screen error are seeing it because the API call to Rowy's server fails with `ERR_SSL_PROTOCOL_ERROR`. The SSL error has been pointed at https://github.com/rowyio/rowy/pull/1352#issuecomment-1648049637 and also over at Rowy's discord. Rowy will be looking at the issue from their backend and infra.

Closing this PR. Thank you everyone.

PR #1368, opened 2023-08-06
Add an accessible resizing option

### LINKED ISSUE #563: Column menu update - Add accessible resize option

nithinrdy:
Fixes #563.

 - Created a new atom to store the table.
 - Width is updated via `setColumnSizing()`.
 - The _"Save column sizes for all users?"_ prompt shows up upon updating the width.
 
Functioning:
- Adds an option to the column context menu -- "Set Column Width". Option is disabled if resizing is disabled.
- Selecting it opens up a modal with a single number input field, pre-filled with the current width of the column.
- Entering a new value and hitting Update resizes the column as needed.

### DIFF
--- a/src/components/ColumnModals/SetColumnWidthModal.tsx
+++ b/src/components/ColumnModals/SetColumnWidthModal.tsx
@@ -1,0 +1,71 @@
+import { useEffect, useState } from "react";
+import { IColumnModalProps } from ".";
+import { selectedCellAtom, tableHeadersAtom } from "@src/atoms/tableScope";
+import { tableScope } from "@src/atoms/tableScope";
+import { useAtom } from "jotai";
+
+import { TextField } from "@mui/material";
+import Modal from "@src/components/Modal";
+import { TableRow } from "@src/types/table";
+import { Header } from "@tanstack/react-table";
+
+export default function ResizeColumnModal({
+  onClose,
+  column,
+}: IColumnModalProps) {
+  const [newWidth, setWidth] = useState<number>(0);
+  const [selectedCell] = useAtom(selectedCellAtom, tableScope);
+  const [tableHeaders] = useAtom(tableHeadersAtom, tableScope);
+  const [selectedHeader, setSelectedHeader] = useState<Header<
+    TableRow,
+    any
+  > | null>(null);
+
+  useEffect(() => {
+    if (selectedCell && tableHeaders) {
+      setSelectedHeader(
+        tableHeaders.find((h) => h.id === selectedCell?.columnKey) ?? null
+      );
+    }
+  }, [selectedCell, tableHeaders]);
+
+  useEffect(() => {
+    selectedHeader && setWidth(selectedHeader.getSize());
+  }, [selectedHeader]);
+
+  return (
+    <Modal
+      onClose={onClose}
+      title="Set Column Width"
+      maxWidth="xs"
+      children={
+        <TextField
+          value={newWidth}
+          autoFocus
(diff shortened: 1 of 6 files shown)


nithinrdy:
Just wondering -- anything else I'd need to do before/while the PR is in review?

il3ven:
Thank you for the PR. The PR works but I feel it can be done in a better way.

In its current form, we can't set size through the keyboard because it uses the mouse event. Another point - calling `setTableHeaders` on every render doesn't feel right.

I propose that we use [`table.setColumnSizing`](https://tanstack.com/table/v8/docs/api/features/column-sizing#setcolumnsizing) function to resize. To access `table` in `setColumnWidthModal` we can store it in an atom. Let me explain it with code.

```Table.tsx```
```js
const [reactTable, setReactTable] = useAtom(reactTableAtom, tableScope)
const table = useReactTable({})
useMemo(() => {
  setReactTable(table)
}, [table])
```

```setColumnWidthModal.tsx```

```js
export default function SetColumnWidthModal({ close }) {
  const [column] = useAtom(columnMenuAtom, tableScope)
  const [table] = useAtom(reactTableAtom, tableScope)
  const onSubmit = (e) => {
    table.setColumnSizing((old) => {
      return {
        ...old,
        [column.name]: e.target.value
      }
    }
  }
  
  return <></>
}
```

Having react table in an atom should be useful in the future too. What do you think about this approach?

src/components/ColumnModals/SetColumnWidthModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+          value={newWidth}
+          autoFocus
+          variant="filled"
+          id="name"
+          label="Width"
+          type="number"
+          fullWidth
+          onChange={(e) => setWidth(Number(e.target.value))}
+          onKeyDown={(e) => {
+            e.key === "Enter" && handleUpdate();
```
nithinrdy, line 48: Tried attaching `handleUpdate` to `onSubmit`, but it didn't work. Resorted to this (`onKeyDown` and then checking for Enter) for now.

Any ideas what I might be missing? I feel like it must be something obvious.

nithinrdy (commented):
Hey @il3ven, thanks a lot for taking a look.

I made some changes based on your suggestions, and I agree its much cleaner and it works just as well, and cutting down on unnecessary updates using `useMemo` is certainly not something I had thought of.

I encountered a small hiccup though, I'd appreciate any ideas. Thanks.

src/components/ColumnModals/SetColumnWidthModal.tsx
il3ven: I have added a form element which handles submission through the button and enter key.

il3ven (approved):
Thank you for the contribution. The code looks good.

src/components/ColumnModals/SetColumnWidthModal.tsx
nithinrdy: Ah right, it wasn't enclosed within a form. _Obvious_ indeed haha 😅
Thanks a lot @il3ven.

PR #1384, opened 2023-08-17
Array column filter

ayusham001:


il3ven (commented):
It looks like this PR has unnecessary changes. You are trying to filter the array on the client's side. It won't work when the array is large. The data can be larger than the available memory and that's why we let firebase handle the filtering and only download the filtered data.

Closing this PR in favour of #1385.

Issue #1382, opened 2023-08-17
FEAT: Array column should be filterable


harinij:
**Is your feature request related to a problem? Please describe.**
Currently, the filter feature does not work for the newly added "Array" column type.

**Describe the solution you'd like**
Allow for the Array column to be filterable.
<img width="647" alt="Screenshot 2023-08-17 at 1 06 22 pm" src="https://github.com/rowyio/rowy/assets/307298/c68f0ec4-43be-4f6c-83f5-2546330a5339">

rishi-raj-jain:
/attempt #1382 (already submitted the PR as above)

PR #1385, opened 2023-08-17
fix: FEAT: Array column should be filterable

rishi-raj-jain:
fix: #1382
/claim #1382

### DIFF
--- a/src/components/fields/Array/Filter.tsx
+++ b/src/components/fields/Array/Filter.tsx
@@ -1,0 +1,24 @@
+import { IFilterOperator } from "@src/components/fields/types";
+
+export const filterOperators: IFilterOperator[] = [
+  {
+    label: "equals",
+    value: "==",
+  },
+  {
+    label: "not equals",
+    value: "!=",
+  },
+  {
+    label: "includes",
+    value: "in",
+  },
+  {
+    label: "not includes",
+    value: "not-in",
+  },
+  {
+    label: "contains atleast one of the following",
+    value: "array-contains-any",
+  },
+];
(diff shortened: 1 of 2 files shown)


rishi-raj-jain:
@harinij @shamsmosowi

Can you take a look? I've extended every method in firestore for arrays: https://firebase.google.com/docs/firestore/query-data/queries#array_membership

Works locally just as fine.

il3ven:
@rishi-raj-jain First of all, thank you for the efforts. Second, the PR needs to be opened against develop. You'll need to rebase. Third, this PR isn't working for all cases. To reproduce, use "not-includes" as the operator.

I believe it isn't working because the translation from `value` to firebase query is missing. It should be done here.

https://github.com/rowyio/rowy/blob/b2983f4db82ebc6c2ad526410a8d20088cc0f245/src/hooks/useFirestoreCollectionWithAtom.ts#L386-L398

rishi-raj-jain:
Thanks for pointing that out.
So I see this only failing for both includes and non-includes. The rest seem to work fine.
By against develop you mean from my develop to rowy's develop, right? Right now, it's from my main to your's develop.

il3ven:
> So I see this only failing for both includes and non-includes. The rest seem to work fine.

Yes

> By against develop you mean from my develop to rowy's develop, right? Right now, it's from my main to your's develop.

The name of your branch can be anything. It doesn't have to be develop. Let's say your branch name is *rishi*. The PR will be from rishi to develop.  However, *rishi* should have all the commits from rowy's develop plus your new commits. Otherwise, there will be merge conflicts which we can see right now.

There are different ways to do this.
- You can create a new branch from develop and cherry-pick the new commits from rishi-raj-jain:main.
- Don't create a new branch but do `git rebase develop` to your current branch.

rishi-raj-jain:
@il3ven 

Thanks for the answers. I do realize it's not the same but the branch. 

I figured out that in and not-in don't work when we want to filter when the filter value is an array itself.

Only, ==, !=, array-contains and array-contains-any are the supported methods as I gather from that link + testing in a simple demo.js file.

rishi-raj-jain:
Creating a fresh PR.

PR #1393, opened 2023-08-28
[BUG] Offline Error Fixed

### LINKED ISSUE #1347: Bug: Offline error

Grenish:
## Issue Fixed
Fixes #1347

/claim #1347

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,0 @@
-yarn lint-staged
--- a/src/components/ErrorFallback.tsx
+++ b/src/components/ErrorFallback.tsx
@@ -20,13 +20,17 @@ export function ErrorFallbackContents({
   resetErrorBoundary,
   ...props
 }: IErrorFallbackProps) {
+  // Detect if the user is offline
   const isOffline = useOffline();
 
-  if ((error as any).code === "permission-denied")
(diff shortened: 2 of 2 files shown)


il3ven:
@Grenish Thank you for giving this a try. Really appreciate it. However, upon deeper inspection it turns out that this issue can't be fixed from the frontend. As it has been noted at https://github.com/rowyio/rowy/pull/1352#issuecomment-1687035489.

PR #1412, opened 2023-09-13
Table name textbox UI error fix

### LINKED ISSUE #1409: Table name textbox UI error on first click
**Describe the bug**
When opening the table settings and trying to edit the Display name, it glitches [see video](https://jam.dev/c/9f03ecaf-be06-4d1f-ab7c-52e35afff3aa). Works after you try to change any other field like description for example.

**Expected behavior**
Display name should be editable in the table settings

AnushDeokar:
closes #1409 
It was watching the wrong field i.e. collection due to which the watchvalue will always be equal to the collection name. 
Also, I have added consistency in the startCase through which if you switch your input focus and comeback again on tablename field the startCase consistency would be maintained.

### DIFF
--- a/src/components/TableSettingsDialog/TableName.tsx
+++ b/src/components/TableSettingsDialog/TableName.tsx
@@ -27,7 +27,7 @@ export default function TableName({ watchedField, ...props }: ITableNameProps) {
         onChange(startCase(watchedValue));
       } else if (typeof value === "string") {
         // otherwise if table name is valid, set watched value to table name
-        onChange(value.trim());
+        onChange(startCase(value.trim()));
       }
     }
   }, [watchedValue, disabled, onChange, value]);
--- a/src/components/TableSettingsDialog/form.tsx
+++ b/src/components/TableSettingsDialog/form.tsx
@@ -213,7 +213,7 @@ export const tableSettings = (
       name: "name",
       label: "Table name",
       required: true,
-      watchedField: "collection",
+      watchedField: "name",
       assistiveText: "User-facing name for this table",
       autoFocus: true,
       gridCols: { xs: 12, sm: 6 },


harinij:
@AnushDeokar the watchvalue is collection as the table name defaults to the collection name with case corrected form if nothing else is mentioned. The bug here is that the Text box was not editable on first attempt however is editable after another field like description is edited.

AnushDeokar:
Hey @harinij, 
The reason why we are not able to change the table name in the first attempt is because in the first attempt onChange(startCase(watchedValue)); is triggered and since the watchValue is set to collection name doing changes in the field is causing it to reset it to collection name. Therefore the tablename field seems to be uneditable in the first attempt. 

However, when we switch field to some other field and comback(let's name it as second attempt) to edit the tablename it works because this time onChange((value.trim()); is triggered which works as expected. 

Another reason why keeping watchfield equal to collection does not seems to be logical because consider the following case:
1. My table name is initially set to let's say Members which is equal to collection name.
2. Now, I want to change my table name to students. I somehow manages to do by changing it the second attempt and I save it.
3. However After saving I realise that I had made a typo mistake I forgot to put an 's' in the end and it's just student. Now I will again go back to table settings and as soon as I try to add an 's' at the end in the table name field(displayed as student) It will get refreshed to 'Members' which is equal to the collection name and again the table name will become uneditable since it's the first attempt. This does not makes sense since I want to start edit from student itself. This is another issue.

Changing the collection name to 'name' will fix this bug. Please let me know if I am wrong at any point. Changes made in this PR works perfectly fine and as expected.

AnushDeokar:
Why did I add a startcase in value.trim()?

Now let's say I am also able to edit the table name in the first attempt. Another change I made is changing onChange(value.trim()) to onChange(startCase(value.trim())) because in the first attempt when I am trying to change the name the case correction is done but when you switch your focus to description and again comback to change tablename the case correction will not be performed. 
You ofcourse cannot notice it right now in the deployment because it's not able to change the tablename in the first attempt.

PR #1437, opened 2023-10-17
ROWY-1542 Add buildship authenticated trigger

bofeiw:


bofeiw:
@shamsmosowi I've added default values for buildship extension to the template.

PR #1443, opened 2023-10-22
Feat: Command Kbar

Hardiegogo:
# Demo :
https://github.com/rowyio/rowy/assets/51075671/09770fe3-8514-46b2-b67d-09fbf794284d

## Features
Actions through kbar and hotkeys

### Navigation:
- Tables page [t]
- Members page [m]
- Settings page [s]
- Tutorial page [j]
- Debug page [d]
- Project settings page [p]

### Specific actions:
- Create table [c]
- Toggle Nav drawer [n]

### Any dependency update?
 Yes, added kbar package.
 
 Tackling #598 in this PR.


shamsmosowi:
Hi @Hardiegogo, thanks for the contribution! this looks like you're on the right track but its only implementing static actions/paths, is it possible to support navigating between tables? Also could you include at least a single use case that performs table level action, such creating column or an extension

Hardiegogo:
@shamsmosowi Let me look into these features, I will start working on them asap.

Issue #1463, opened 2023-11-02
Bug with the bulk selection column with freeze columns


harinij:
**Describe the bug**
Bulk selection checkboxes option column is not fixed and it causes issues when horizontal scrolling with freezed columns. See video for explanation.

**To Reproduce**
See the recording
https://github.com/rowyio/rowy/assets/307298/7f6fda8a-9081-4a4d-88ae-01877d55bfbc

**Expected behavior**
Bulk selection options column should be fixed in the table.

janvi01:
@harinij I would like to work on this. Please assign me.

AnushDeokar:
/attempt #1463

Hey @harinij @janvi01, By looking at the previous issues I noticed that issues with bounties are not assigned and they are given on first successful attempt basis so as I started working on it as soon as it came without asking to assign. If this is not the case then kindly ignore my attempt. Thanks.

janvi01:
@harinij I submitted a PR. Thanks for assigning me.

il3ven:
@harinij I have merged https://github.com/rowyio/rowy/pull/1466 and closed others. You can close the issue and pay the bounty.

PR #1472, opened 2023-11-04
fix: bulk selection column not fixed

rad10wave:
/claim https://github.com/rowyio/rowy/issues/1463
Fixes https://github.com/rowyio/rowy/issues/1463
Added fixed attribute to the Checkbox column


il3ven:
Thanks for giving the fix a try. However, it doesn't work for me. Also, a fix for this has been merged.

Issue #1478, opened 2023-11-10
[UI] [BUG] - Filters in sub-table keeps refreshing


Chadha93:
**Describe the bug**
When adding/removing a Filter to a sub-table, the page is getting into loops and refreshes. It is not updating the filter value on Firestore.

https://github.com/rowyio/rowy/assets/65453826/6f06ff7f-ee98-45f6-9854-6c8d9688899b

**To Reproduce**
1. Create a table - https://docs.rowy.io/how-to/create-table
2. Add a suitable - https://docs.rowy.io/field-types/array-subtable#creating-an-array-sub-table
3. Add and remove filter in the sub-table. 

**Expected behavior**
On addition or removal of any filter it should also update the filter value from the Firestore array. Which should then update the value in Rowy sub-table filters. 

**Rowy Run version**
Rowy - V3.0.0

Wr4th100:
@harinij, the PR got merged, could you please close this issue?

harinij:
Thanks @Wr4th100 for the contributions. Closing this now.

PR #1482, opened 2023-11-12
fix: subtable filters refresh bug

Wr4th100:
/claim #1478

Solves #1478 

https://github.com/rowyio/rowy/assets/68044488/b030a20f-c0f1-49c1-9fcb-6b1ad9b13045


il3ven (approved):
Awesome, thanks for the PR.

Issue #1483, opened 2023-11-15
BUG: Filter on the table works, but the same applied to the Table Filter doesn't


harinij:
**Describe the bug**
When you filter by a value in the table it works as expected. However if you try to apply the same filter at table level to all users using that table as an ADMIN user - it does not work.

**To Reproduce**
See video

https://github.com/rowyio/rowy/assets/307298/7a57493c-958f-41aa-9261-378e2d162716

ezhil56x:
Hello @harinij 👋🏻,
I would like to work on this issue!

janvi01:
@harinij I am not able to reproduce this issue, the case mentioned works fine.

ezhil56x:
@harinij I have setup my local dev env and I'm unable to reproduce this issue


https://github.com/rowyio/rowy/assets/103899034/f0f40c5e-60c8-4e81-954e-13347b3fe47b

zeeshan1212:
> 💡 @janvi01 submitted a [pull request](https://github.com/rowyio/rowy/pull/1504) that claims the bounty. You can [visit your bounty board](https://console.algora.io/org/rowyio/bounties) to reward.

@janvi01 I've already submitted pull request for this and worked on this.

zeeshan1212:
@harinij the PR got merged, could you please close this issue?

PR #1486, opened 2023-11-16
fix: #1483 Applying filters using table

zeeshan1212:
/claim #1483

### DIFF
--- a/src/components/TableToolbar/Filters/FilterInputs.tsx
+++ b/src/components/TableToolbar/Filters/FilterInputs.tsx
@@ -165,6 +165,7 @@ export default function FilterInputs({
                       column: selectedColumn,
                       _rowy_ref: {},
                       value: query.value,
+                      onSubmit: () => {},
                       onChange: (value: any) => {
                         const newQuery = {
                           ...query,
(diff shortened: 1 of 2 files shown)


il3ven:
@zeeshan1212 Thank you very much for the contribution. The PR does fix the bug. Instead of using default input for `Number_` component can you pass in `onSubmit` from the filters component?

https://github.com/rowyio/rowy/blob/80b2f0fbe7551b3184a30574d70567292acb5c75/src/components/TableToolbar/Filters/FilterInputs.tsx#L157-L179

In the above file, we can write `onSubmit = () => {}` like `onChange` and it should fix the bug

Reasons for this change:
- Default input fixes this bug for `Number` input field but the bug can be still be there for other input types.

zeeshan1212:
> @zeeshan1212 Thank you very much for the contribution. The PR does fix the bug. Instead of using default input for `Number_` component can you pass in `onSubmit` from the filters component?
> 
> https://github.com/rowyio/rowy/blob/80b2f0fbe7551b3184a30574d70567292acb5c75/src/components/TableToolbar/Filters/FilterInputs.tsx#L157-L179
> 
> In the above file, we can write `onSubmit = () => {}` like `onChange` and it should fix the bug
> 
> Reasons for this change:
> 
> * Default input fixes this bug for `Number` input field but the bug can be still be there for other input types.

yeah sure

zeeshan1212:
@il3ven changes addressed! please check
https://github.com/rowyio/rowy/pull/1486/commits/3ec15f0121d3abba45a6a3cb05e839122830c80c

il3ven (approved):

PR #1504, opened 2023-12-12
fix: Applying filter using Table filter

### LINKED ISSUE #1483: BUG: Filter on the table works, but the same applied to the Table Filter doesn't

janvi01:
The PR fixes #1483 

/claim #1483


janvi01:
@il3ven Kindly review. Thanks.

il3ven:
I have merged #1486 as that was opened before this.

Issue #1531, opened 2024-01-26
Animation to search tables gets slower as number of table grows


bofeiw:
If there are a lot of tables in the project, searching for a table will animate slower and the table takes longer.

Example in demo project: searching for "story" will take animate a bit.

https://github.com/rowyio/rowy/assets/34177142/75b177aa-d081-478a-94c6-e3909618222c

An extreme example, there are 170 tables, search took 6 seconds to animate.

https://github.com/rowyio/rowy/assets/34177142/fb9379b8-b32c-4f63-9315-a4e1df0950eb

**To Reproduce**
1. Go to https://demo.rowy.io/tables
2. Search for "story"
3. Notice the animation is slow

**Expected behavior**
Search table animation should always be fast regardless of number of tables.

mehulmathur16:
/attempt #1531 Can this issue be assigned to me ?

PR #1534, opened 2024-02-01
fix: search table result loading

janvi01:
/claim https://github.com/rowyio/rowy/issues/1531

Fix - Animation to search tables loading

Implemented -

1. Decreased the values used in the timeout prop for the SlideTransition components.
2. Avoid unnecessary rendering when the user types aggressively, using the debounce technique.

Demo -

https://github.com/rowyio/rowy/assets/34344234/dd3bc50e-0f4b-4374-a861-f9ff72ee5a84


janvi01:
@bofeiw @harinij Kindly review, Thanks.

Issue #1542, opened 2024-02-24
Bulk edit support - edit a number of rows at once


bofeiw:
**Is your feature request related to a problem? Please describe.**
Ability to edit multiple rows at the same time to improve efficiency.

**Describe the solution you'd like**
Solution 1: Rowy already support multi row selection. We can add a button "bulk edit" when multiple rows are selected.
<img width="596" alt="Screenshot 2024-02-24 at 16 39 30" src="https://github.com/rowyio/rowy/assets/34177142/b36b199f-1e70-4e85-9e8e-7ccd5f178585">

Solution 2: Like Google Sheets, we can allow a cell's corner to be draggable to copy the value to other cells.
<img width="183" alt="Screenshot 2024-02-24 at 16 40 31" src="https://github.com/rowyio/rowy/assets/34177142/ec321bb6-e84f-45f0-bb10-c9dcc56979c9">

**Describe alternatives you've considered**
Currently user can copy the value and manually paste to other cells as a workaround. However this is not efficient.

AnushDeokar:
Hey @bofeiw, Do you mind if I give this a go? I was thinking to implement this (Solution 1) similar to how we have Filter option.

PR #1547, opened 2024-03-03
Add Support for Bulk Edit

AnushDeokar:
enhancement #1542 

This PR attempts to add the support for Bulk Edition of Rows. The implementation in terms of design is similar to that of Filter option. I have created a new atom to store the bulk edit popover state. I am filtering out the column types which can be editable and then handling the user change in the query state. Then the bulk update is done asynchronously by calling the updateField method.

  

https://github.com/rowyio/rowy/assets/58873530/56c0ecd9-15e5-48bd-8d22-857ca00e7521

### DIFF
--- a/src/atoms/tableScope/ui.ts
+++ b/src/atoms/tableScope/ui.ts
@@ -2,7 +2,11 @@ import { atom } from "jotai";
 import { atomWithStorage, atomWithHash } from "jotai/utils";
 
 import type { PopoverProps } from "@mui/material";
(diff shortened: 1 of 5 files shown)


AnushDeokar:
@bofeiw, Could you please have a look at this? I'm still refining it more, but I'd appreciate confirming whether my grasp of this enhancement aligns with the intentions or not. If It could be done in any better way let me know. Thanks

PR #1595, opened 2026-05-19
Fix filtered table rows including local-only rows

### LINKED ISSUE #1527: Out of order rows do not work with sorts/filters
**Describe the bug**

If there are out of order rows, they are always displayed regardless of the sorts or filters.

Here is an example screenshot of the issue. Filter is set to percentage=0.1 however there are rows whose percentage is 0 appearing in the UI. Notice that the row counter displays "loaded all 48 of 13 rows" which is incorrect.

Screenshot 2024-01-22 at 07 06 22

The cells with dotted red lines are the columns that require a value but those rows do not have a value. They are the "out of order" rows, and with the current code logic, those out of order rows have a high priority for display.

https://github.com/rowyio/rowy/blob/0099bd1daca3f722b05ebbd5294eb31837382b2d/src/atoms/tableScope/table.ts#L171
```js
export const tableRowsAtom = atom<TableRow[]>((get) => {
  const rowsDb = get(tableRowsDbAtom);
  const rowsLocal = get(tableRowsLocalAtom);
   
  ...

  // Merge the two arrays
  return [
    ...rowsLocalToMerge,
    ...rowsDb.filter((row) => rowsDbMap.has(row._rowy_ref.path)),
  ];
});
```

`tableRowsDbAtom` contains rows that are filtered, `tableRowsLocalAtom` contains rows that are out of order. `tableRowsAtom` combines the two row sets and the table UI directly renders them. This causes the our of order rows to always display in the table UI.

The "loaded all 48 of 13 rows" issue with the row counter is also related here. 48 is the combined number of rows (`tableRowsAtom`), 13 is the correct Firebase server count rows with filters.

**To Reproduce**
1. In any table, have some columns and rows
2. Add a new short text column and leave rows empty, mark it as required
3. Apply filters, all rows are always displayed despite what filters are

https://github.com/rowyio/rowy/discussions/1315 is an example of this issue.

**Expected behavior**
Out of order rows should work with filters/sorts.

**Possible solutions**

1. When sorts/filters are applied, do not include any out of order rows in `tableRowsAtom`. This is the simplest solution.
2. Display out of order rows in a separate section. Need to think about how to implement the separation.
3. Redesign how row atoms are structured so out of order rows are also filtered and sorted.

dttxorg:
Fixes #1527

## Summary
- Keep merging local edits into rows that are returned by the server query.
- Hide local-only out-of-order rows while filters or sorts are active, so filtered/sorted tables only show rows from the server result set.
- Leave the unfiltered and unsorted merge behavior unchanged.

## Notes
This is intentionally a focused data-merging fix with no UI or navigation changes. It addresses the $50 bounty request in #1527.

## Verification
- `yarn lint-staged`
- `yarn vitest run src/atoms/tableScope/rowActions.test.ts`
- `yarn tsc --noEmit`


dttxorg:
Note: the two Vercel status checks currently fail with "Authorization required to deploy" and link to Vercel authorization pages rather than build logs from this branch. I ran the local lint-staged, focused Vitest file, and TypeScript checks listed in the PR description.

PR #1598, opened 2026-05-25
Fix out-of-order local rows when filters/sorts are active

TheDerbiedOne:
/claim #1527

Fixes a case where out-of-order (local) rows bypass active filters/sorts and also inflate the “loaded all X of Y rows” counter.

Approach:
- When filters or sorts are active, `tableRowsAtom` no longer surfaces local out-of-order rows that aren’t in the filtered DB result set.
- Local edits for rows that *are* in the DB result set are still merged, preserving UI correctness without breaking filtering/counts.

Tests:
- Added `src/atoms/tableScope/tableRowsAtom.test.ts` covering the filtered vs unfiltered behavior.

Notes:
- I followed the issue’s “Possible solution #1” (exclude out-of-order rows when filters/sorts are applied).


TheDerbiedOne:
Note: the only reported check failures on this PR are Vercel deploy jobs that require org authorization (“Authorization required to deploy.”).

No further changes on my side unless you’d like anything adjusted in the patch itself; happy to iterate if review feedback comes in.

